PR #123, opened 2026-04-14
Fix unstaged rename parsing in GitStatusParser

### LINKED ISSUE #122: Handle unstaged renames and copies in Git status parser
### Problem

`GitStatusParser.parseStatusPorcelain` only treats rename/copy entries as two-token records when the staged status (`X`) is `R` or `C`. Git also emits renames/copies in the unstaged column (`Y`), for example `" R old.swift\0new.swift\0"` for an unstaged rename.

In that case Muxy currently parses the old path as the current path, drops `oldPath`, and loses the rename/copy metadata for the file list.

### Expected

The parser should treat either status column reporting `R` or `C` as a rename/copy entry and use the second token as the current path.

### Notes

This is reproducible from the parser alone with a porcelain sample like:

```
 R old.swift

txhno:
## Summary
- handle rename/copy porcelain entries when either Git status column reports R/C
- add coverage for an unstaged rename record so the parser keeps the new path and oldPath metadata

## Validation
- `git diff --check`
- `swift test` not run in this environment: Fedora host, no Swift toolchain, package targets macOS 14+

Closes #122

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Setup GhosttyKit
         run: scripts/setup.sh
         env:
-          GH_TOKEN: ${{ secrets.GH_PAT }}
+          GH_TOKEN: ${{ github.token }}
 
       - name: Build
         run: swift build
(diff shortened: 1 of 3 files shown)


saeedvaziry:
Fixing the checks...

PR #140, opened 2026-04-15
Add configurable quick-open editor

### LINKED ISSUE #139: Editor tabs use preferred editor?
### Problem

Really love what you're doing with this the best project focus terminal i've seen love how minimal you're keeping it. 

One feature request I would have, instead of editor tabs being built-in editor, which I think is useful, would it be possible to have a drop-down to allow a user to pick their own preferred editor? So, for example, in my case it would be `nvim`, and we would just launch a tab pointing at the file in question with `nvim`?

### Proposed Solution

- select preferred editor in preferences.
- when picking a file to edit, a new tab is created with the preferred editor. 

### Alternatives Considered

_No response_

### Additional Context

_No response_

jkp:
## Summary
- add a quick-open editor preference with built-in editor and terminal command modes
- launch terminal editor tabs directly through Ghostty startup commands and close them when the editor exits
- hide built-in editor-only settings when terminal command mode is selected

Closes #139

## Validation
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer mise exec -- scripts/checks.sh --fix`

### DIFF
--- a/Muxy/Models/AppState.swift
+++ b/Muxy/Models/AppState.swift
@@ -24,6 +24,7 @@ final class AppState {
         case createTab(projectID: UUID, areaID: UUID?)
         case createVCSTab(projectID: UUID, areaID: UUID?)
         case createEditorTab(projectID: UUID, areaID: UUID?, filePath: String)
+        case createExternalEditorTab(projectID: UUID, areaID: UUID?, filePath: String, command: String)
         case closeTab(projectID: UUID, areaID: UUID, tabID: UUID)
         case selectTab(projectID: UUID, areaID: UUID, tabID: UUID)
         case selectTabByIndex(projectID: UUID, areaID: UUID?, index: Int)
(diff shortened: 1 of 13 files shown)


saeedvaziry:
This is cool!

saeedvaziry:
Going to make some changes. Instead of quick open editor. this should be default editor, cause on the VCS you can also open a file and this can be used there too

jkp:
will it still work with command-p? thats the main win for me...i can obv open anything in a terminal already but being able to quickly find a file from anywhere and get an editor pane is so useful. 

on a philosophical note. im not sure what your vision for the app is... was using `cmux` but i didnt like a bunch of things there and it was a serious resource hog over time. it literally killed my battery in hours. 

what i love about what you've build here is the it plays to the "i live in the terminal and spend most of my day in coding assistants" but without loads of fancy machinery on top. it looks and feels great. 

if you are open to it i think there are a few usability features that would take this to another level for a keyboard warrior like me. eg quick copy (like what `wezterm` has)... if it were me id just make all these things additive and optional so the thing doesnt get bloated like it seems `cmux` has. just my two cents...

i tried all the offerings on the awesome libghostty list and what you built is fas and away the nicest imho. well done!

saeedvaziry:
> will it still work with command-p?

Yes of course

> on a philosophical note. im not sure what your vision for the app is

Not a vision but the direction I'd like to keep:

1. Remaining lightweight. low bundle size. less possible memory and CPU usage!
2. The app is for terminal users! so no diversion to fancy desktop stuff, but in the meantime, more visibility while you work in terminal like VCS, Vertical tabs
3. Keyboard first
4. Integration with TUIs for better notification handling (standards and some custom cases like opencode and claude)
5. Portability, I am on terminal but I want to stay on terminal even if I leave my laptop (cooking an idea of a mobile app)

saeedvaziry:
Ideas are always welcome

jkp:
ok well keyboard first is why i got here! im all about that. patch incoming. lets see what you think.

PR #142, opened 2026-04-15
Add terminal quick select

jkp:
## Summary

Adds keyboard-driven Quick Select for terminal panes.

- Detects URLs, paths, git SHAs, UUIDs, IP addresses, and email addresses in visible terminal text
- Labels matches so typing the label copies the match, with Shift on the final label key copying and pasting
- Adds QWERTY, Colemak, Colemak-DH, Dvorak, Workman, and custom home-row label layouts in Terminal settings
- Adds Quick Select to the configurable shortcut system with a default `Ctrl+Shift+Space` binding
- Documents the feature in the README and architecture map

## Motivation

The goal is a keyboard-first selection flow for terminal text. The interaction is inspired by Vim/Vimari-style hint labels: enter a mode, type a short label, and act on the target without reaching for the mouse. WezTerm has a similar quick-select workflow for URLs and other terminal text, and this brings that kind of terminal-native affordance into Muxy.

This is intentionally a first step toward richer keyboard-driven terminal selection. Quick Select handles the high-value case first: copying or pasting visible URLs, paths, commits, and addresses directly from the terminal using only the keyboard.

## Validation

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer scripts/checks.sh --fix`

### DIFF
--- a/Muxy/Commands/MuxyCommands.swift
+++ b/Muxy/Commands/MuxyCommands.swift
@@ -54,6 +54,12 @@ struct MuxyCommands: Commands {
                 NotificationCenter.default.post(name: .findInTerminal, object: nil)
             }
             .shortcut(for: .findInTerminal, store: keyBindings)
+
+            Button("Quick Select") {
+                guard isMainWindowFocused else { return }
+                NotificationCenter.default.post(name: .quickSelect, object: nil)
+            }
+            .shortcut(for: .quickSelect, store: keyBindings)
         }
 
         CommandGroup(replacing: .newItem) {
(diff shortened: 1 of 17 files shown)


jkp:
image

jkp:
not at all precious about the implementation - i just threw this together quickly to see if it would be feasible. i think this plus a "copy/visual mode" would be amazing for keyboard first.  i started here as this is far easier and covers a lot of cases.

saeedvaziry:
I am trying to understand this feature. what exactly (technically) I should expect it to find text?

image


Is this expected?

saeedvaziry:
I was thinking something like Alacritty for Ctrl+Shift+Space

video

jkp:
the implementation might have gaps and need some polish it should use regex to allow quick copy of matches it finds... but something like the alacrity solution is what id like to see as a more general solution.  wezterm has both and i found them to be quite complementary as many times you just want to copy a nearby url or file path...thats what this feature is supposed to do :)

https://wezterm.org/copymode.html
https://wezterm.org/quickselect.html

saeedvaziry:
I think I can get onboarded with the QuickSelect mode. tho I need to install Wezterm and see what I should expect from it

saeedvaziry:
@jkp would you be able to resolve the conflicts? or feel free to close and open a new one

PR #156, opened 2026-04-16
Add manual import/refresh support for existing Git worktrees

jkp:
## Summary

Add manual import/refresh support for existing Git worktrees in the sidebar.

Muxy can already create Git worktrees, but those worktrees were effectively Muxy-specific because the sidebar only showed worktrees already persisted by the app. This change lets Muxy recognize worktrees that were created outside Muxy and show them under the owning project in the sidebar.

The main motivation is interoperability with external worktree workflows, especially worktrees created in other environments like Claude Code or Codex.

## Design Rationale

This is intentionally the most minimal version of the feature.

I wanted the smallest, least invasive change that makes external worktrees usable in Muxy without changing ownership semantics or adding background behavior. Concretely, that means:

- manual refresh instead of timers or automatic polling
- no background cleanup of externally discovered worktrees
- imported external worktrees are treated as known by Muxy, but not owned by Muxy
- existing Muxy-managed worktree creation and removal behavior stays intact

That keeps the implementation predictable and low-risk:

- no periodic scans that could introduce performance overhead
- no destructive cleanup of worktrees Muxy did not create
- no broader architectural changes just to support cross-environment compatibility

This gives users the immediate benefit that they can create worktrees however they want, wherever they want, and still have those worktrees appear in Muxy for organization and navigation.

There is room to do more later, but this first step is deliberately conservative.

## What Changed

- added a manual Refresh Worktrees action in the sidebar worktree UI
- merged git worktree list --porcelain results into WorktreeStore
- preserved existing Muxy worktree IDs by matching on path, so saved workspace state stays attached
- marked imported worktrees as external
- blocked destructive removal for external worktrees
- added workspace restore fallback by worktreePath when worktreeID no longer matches
- updated architecture docs

## Testing

- added store tests for importing external worktrees, preserving IDs by path, and retaining absent stored entries
- added restore tests for worktreePath fallback
- verified focused worktree/workspace test suites pass


saeedvaziry:
image
why it created this?

jkp:
Thanks for the review — addressed each point below.

### @saeedvaziry — "why it created this?"
That duplicate `feat/worktree-refresh` row in the popover was a path-mismatch bug in `refreshFromGit`: it compared the persisted `project.path` against the path returned by `git worktree list --porcelain` using string equality. When the project was opened via a symlinked path (or the stored path differed from the canonical path git reports), the primary worktree didn't match and got imported a second time as if it were a separate external worktree.

Fixed by canonicalizing paths with `URL.standardizedFileURL.resolvingSymlinksInPath().path` before comparing. Covered with a new test (`refreshFromGit treats symlinked primary paths as the primary worktree`).

### Copilot review
1. **`Dictionary(uniqueKeysWithValues:)` can trap on duplicate paths** — Replaced with an iterative builder that tolerates duplicates and prefers the primary entry on collision. New test: `refreshFromGit tolerates duplicate persisted paths without trapping`.
2. **External worktrees still destructively cleaned up by other paths** — Good catch. `cleanupOnDisk(worktree:repoPath:)` now no-ops when `worktree.isExternallyManaged`, and `cleanupOnDisk(for:knownWorktrees:)` filters external worktrees out before invoking git/filesystem removal. This covers project removal, post-merge cleanup, and the direct remove paths. The UI already hides the Remove button for externals; this is the backend invariant that guarantees it.
3. **`PopoverFooterAction` UUID regenerated on every recompute** — Replaced with a stable id derived from `title|icon` (overridable via initializer).
4. **VCSTabView refresh silently swallows errors via `try?`** — Routed through the shared `WorktreeRefreshHelper` so it uses the same alert surface as the sidebar.
5. **Duplicated `refreshWorktrees()`/`presentRefreshError()` in ProjectRow + ExpandedProjectRow** — Extracted to `WorktreeRefreshHelper` (a `@MainActor` enum). Both sidebar rows and the VCS tab now share one implementation; each view passes a `Binding<Bool>` for its local refresh state.

Architecture doc updated to mention the symlink-resolution behavior and the cleanup invariant for externally discovered worktrees.

saeedvaziry:
@jkp could you please upload a screen recording of the feature? Thanks

jkp:
https://github.com/user-attachments/assets/c0f961ea-f849-4106-b810-d339b0729d58

Issue #164, opened 2026-04-17
Moving Muxy from external monitor to Mac Retina display results in inability to scroll to bottom of content


unstatusthequo:
### Description

Opened Muxy on ultrawide monitor. Adjusted text size a bit, though I don't think that is the root cause of the issue, which is when moving muxy to the Retina laptop screen, there is what looks like a resolution issue where the bottom of the terminal session is cutoff. Scrolling does not fix. oddly, squishing the window vertically eventually shows the bottom (Coding agent prompt area). Moving the window back to ultrawide restores full view. 

### Steps to Reproduce

See description. 

### Muxy Version

0.17.0

### macOS Version

26.4.1

saeedvaziry:
I don't have this setup to fix this issue. I'd ask you or anyone who has the setup to reproduce and fix and contribute

PR #170, opened 2026-04-18
Fix terminal content cutoff when moving window between displays

RanaMoizHaider:
## Summary
This PR fixes issue #164. Terminal content was not showing properly when window was moved between different displays.

## Changes

- Added a small delayed resize to let AppKit settle when the window moves between displays
- Updated the view/layer scaling step to match the new display first, then resize the Ghosty surface
- Hooked into screen/backing change notifications to re-apply safe-area neutralization, which was getting out of sync after moving between monitors.
- Switched backing pixel rounding from `ceil` to `floor` to avoid occasional off-by-one sizing that could clip the bottom row.

## Testing

### DIFF
--- a/Muxy/MuxyApp.swift
+++ b/Muxy/MuxyApp.swift
@@ -309,6 +309,8 @@ struct WindowConfigurator: NSViewRepresentable {
             let names: [Notification.Name] = [
                 NSWindow.didResizeNotification,
                 NSWindow.didEndLiveResizeNotification,
+                NSWindow.didChangeScreenNotification,
+                NSWindow.didChangeBackingPropertiesNotification,
                 NSWindow.didExitFullScreenNotification,
                 NSWindow.didEnterFullScreenNotification,
             ]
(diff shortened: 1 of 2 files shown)


saeedvaziry:
Thanks @RanaMoizHaider for the contribution, would you share a screen recording or screenshot of the issue and the fix? I can't reproduce myself due to setup. Thanks

RanaMoizHaider:
Here are a few Before shots:

https://github.com/user-attachments/assets/8eea4c5f-2b05-45e6-bbc7-4d4d05e53857


https://github.com/user-attachments/assets/9653447b-beb6-4fe3-ae8f-cfa71da49b20

And here is the After shot:

https://github.com/user-attachments/assets/6914844a-0078-4721-805d-ebc6c8ba5416

Couldn't record two screens at the same time with native video recorder so using OBS. I don't know how to use it so sorry for the quality. I hope these help.

PR #205, opened 2026-04-21
Remove `setAppIcon()` dock icon override

pkgamma:
## Summary

In a early commit, I believe `setAppIcon()` was added to handle icon problem at that time, but now it's causing an issue where after app launches, the icon is programmatically replaced and user's custom icon (whether applied through Finder Inspect, an app like Pictogram, or even direct replacement of `AppIcon.icns` inside .app bundle) is not respected, and icon always changes to default Muxy icon.

There is an `setAppIcon()` method that does this, in there the runtime `applicationIconImage` prop always takes precedence over what macOS reads from .icns bundle resource, which destroys icon customization I have made, which is bad UX.
  
Since macOS automatically uses `AppIcon.icns` from the app bundle as the dock icon, I removed the `setAppIcon()` call completely, and I've made sure same icon is still shown, just that user now have the freedom to set what they want!

## Changes

- removed `setAppIcon()` in `MuxyApp.swift`

## Testing

## Screenshots

image

### DIFF
--- a/Muxy/MuxyApp.swift
+++ b/Muxy/MuxyApp.swift
@@ -114,7 +114,6 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     func applicationDidFinishLaunching(_ notification: Notification) {
         NSApp.setActivationPolicy(.regular)
         NSApp.activate()
-        setAppIcon()
         _ = GhosttyService.shared
         ThemeService.shared.applyDefaultThemeIfNeeded()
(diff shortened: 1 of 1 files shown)


saeedvaziry:
@pkgamma It was there for a reason, to have app icon on the dev mode. Can you adjust it to work for both purposes?

pkgamma:
Weirdly, this doesn't reproduce anymore in latest build, so closing for now!

PR #239, opened 2026-04-25
Feature emoji for project

IlyaRastorguev:
## Add emoji as project icon

Users can now set any emoji as a project icon via the context menu in the sidebar.

## What changed

**New: `EmojiPickerController`**
Singleton that wraps macOS character palette (`NSApp.orderFrontCharacterPalette`) behind a hidden 1×1 `NSPanel`. The panel becomes key responder to intercept `insertText` and capture the selected emoji, then dismisses immediately. `prepare()` warms the input context on sidebar appear to reduce first-open latency.

**`ProjectStore.setIcon(id:to:)`**
New persistence method — stores emoji string on `Project.icon`, follows same pattern as `setIconColor`.

**Sidebar context menu**
Both `ProjectRow` and `ExpandedProjectRow` gain two new menu items: "Set Emoji…" (opens picker) and "Remove Emoji" (shown only when emoji is set).

**Icon render priority**
Emoji takes precedence over logo image, which takes precedence over the default letter tile. Background fill is suppressed when emoji is active.

## Testing

`ProjectStoreTests` — covers `setIcon` for both set and clear cases using a stub persistence layer.

## Screenshots

https://github.com/user-attachments/assets/846d12ca-bd10-475d-81ed-119b7d6a4949


saeedvaziry:
Not needed feature. it just adds more code to maintain

PR #240, opened 2026-04-25
Feature vibrant bg

IlyaRastorguev:
## Add vibrant/transparent window background

Adds user-controlled background transparency and blur for the main window, letting users see content behind Muxy through a native macOS vibrancy effect.

### Changes

**`AppAppearancePreferences` (new)**
Namespace for all appearance persistence logic. Reads/writes `background-opacity` and `background-blur` to the Ghostty config, clamps values to valid ranges, and caps opacity at 0.85 when blur is active so the vibrancy effect remains visible.

**`WindowConfigurator`**
Receives `transparencyLevel` and `blurRadius` as props and applies them to the `NSWindow` on every config change. When blur > 0: sets window non-opaque, inserts an `NSVisualEffectView` (`.behindWindow`, `.underWindowBackground`) pinned to the content view, and calls `ghostty_set_window_background_blur`. When blur == 0: removes the effect view and restores an opaque window.

**`MuxyTheme`**
Cache invalidation now includes opacity and blur radius so theme snapshots regenerate when appearance settings change. Background color alpha is derived from `AppAppearancePreferences.backgroundOpacity`.

**`AppearanceSettingsView`**
Two new sliders — Transparency (0–100%) and Blur (0–64 px) — backed by `@AppStorage`. Changes write to Ghostty config and trigger a config reload immediately.

**`MuxyConfig`**
New `removeConfigValue(_:)` method used to purge the legacy `background-blur-radius` key on first write.

## Testing

`AppAppearancePreferencesTests` covers clamping, config value formatting, and the blur/opacity interaction rule.

## Screenshots

https://github.com/user-attachments/assets/ea381e07-3276-4df2-8789-3290d4c55b33


saeedvaziry:
image

This is not looking good. also causing lag in the UI. I wouldn't sacrify the performance for it

PR #245, opened 2026-04-25
Add `Install CLI` to open projects in muxy from teminal `muxy /path/to/project`

RanaMoizHaider:
## Summary
Adds an "Install CLI" menu action that copies the bundled muxy CLI wrapper script to `/usr/local/bin` (with `~/bin` fallback), enabling users to open projects from the terminal with `muxy .` or `muxy /path/to/project`.

## Changes

- Added installCLI() to CLIAccessor — copies the CLI script to /usr/local/bin/muxy, asking for admin privileges via AppleScript when needed
- Registered muxy:// URL scheme in Info.plist for production use
- Added application(_:open:urls:) handler in AppDelegate to open projects from URLs (muxy: scheme, file:// URLs via open -b)
- Extended NotificationSocketServer to handle open-project messages via Unix socket — reliable fallback for development mode
- Added openProjectFromPath() to CLIAccessor that adds or activates a project by path

## Testing

- [x] `scripts/checks.sh` passes (formatting, linting, macOS build all passed. iOS build failed because I do not have iOS simulator installed)


saeedvaziry:
Good one!

saeedvaziry:
@RanaMoizHaider Would this work if I have the Muxy open and run this?

I think would be nice to have this behaviour:

1. I am in muxy in a project, I run `cd to-some-dir-outside-of-the-current-project`
2. then run `muxy .` and it will just attach that folder to the sidebar of the currently open muxy not a new Muxy instance

Would be useful

RanaMoizHaider:
That would indeed be nice. I'll look into it

PR #248, opened 2026-04-26
add select menu to open terminal with a cli

eq0:
add select menu to open terminal with a cli, cause it's annoying to write command everytime we open the terminal

## Summary

this PR does add select menu to choose either gemini cli, claude code, or normal terminal

## Testing

## Screenshots

image


saeedvaziry:
Thanks @eq0 for the PR.

I don't think this feature is needed for Muxy as it just brings maintenance. there are easier ways to do this like setting aliases on your bash.

Also, many uses claude in dangerously-skip-permission mode, and supporting all of these will be just maintenance overhead.

Closing this PR in favor of keeping the app simple with less maintenance possible

PR #304, opened 2026-05-01
Update sidebar UX for better overview of active worktrees and projects

iodic:
## Summary

This changeset updates the projects & worktree styling in the sidebar to communicate the structure and highlight the active project more clearly.

## Changes

- Remove worktree checkmarks
- Remove selected-worktree background highlight for inactive projects
- Remove persistent background highlight on inactive project headers
- Push sidebar text to full foreground color to increase contrast
- Brighten default monogram chip background
- Set bold worktree text to selected worktree of active project
- Widen sidebar to 290pt
- Bump font sizes and worktree row spacing

## Testing

### Before

Screenshot 2026-05-01 at 00 09 23

### After

Screenshot 2026-05-01 at 00 19 44

### DIFF
--- a/Muxy/Views/Sidebar.swift
+++ b/Muxy/Views/Sidebar.swift
@@ -2,7 +2,7 @@ import SwiftUI
 
 enum SidebarLayout {
     static let collapsedWidth: CGFloat = 44
-    static let expandedWidth: CGFloat = 220
+    static let expandedWidth: CGFloat = 290
     static let width: CGFloat = 44
 
     static func resolvedWidth(
--- a/Muxy/Views/Sidebar.swift
+++ b/Muxy/Views/Sidebar.swift
@@ -2,7 +2,7 @@ import SwiftUI
 
 enum SidebarLayout {
     static let collapsedWidth: CGFloat = 44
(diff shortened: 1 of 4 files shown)


saeedvaziry:
It looks good, however I'd revert the sidebar width. its too wide now and taking a lot of space

saeedvaziry:
image

And since we're doing this, I think would be nice if we make this default worktree to look like the others WDYT?

iodic:
@saeedvaziry, I have pushed the sidebar width update. Restored it back to 220pts.

Looking into updating the primary worktree view now. Is this what you had in mind:

Screenshot 2026-05-01 at 23 54 17

I think it looks good. (I kinda liked that `primary` label but it's likely not that informartive/useful in the end.)

saeedvaziry:
We can still have the primary badge there. What I am mostly trying to achive is to make the height similar to other worktrees.
Showing project name there is redundant.
it can be the branch name and the primary badge in front of it.

iodic:
Here is the updated view:

Screenshot 2026-05-02 at 00 24 34

And the worktree selector:

Screenshot 2026-05-02 at 00 25 13

PR #308, opened 2026-05-01
Add project path sync and UI shortcut improvements

dtateks:
## Summary
- Adds app-wide themed quick tooltips with a custom SwiftUI/AppKit hover tracker, faster display timing, and cleanup when opening popovers or actions.
- Adds smoother sidebar drag resizing with persisted custom width, auto collapse/expand thresholds, and a setting to hide top-bar navigation arrows.
- Adds project commands for Cmd-N duplicate project, Cmd-Shift-N new window, command-menu project rename, special-key shortcut support, and single-terminal CWD-to-project path sync.

## Features
- New Project command duplicates the active project and inserts it next to the current project for faster multi-window/project workflows.
- New Window command opens another main Muxy window from the app command menu and Cmd-Shift-N.
- Rename Project command starts inline rename on the active sidebar project row from the command menu.
- Single-terminal CWD sync updates ProjectStore path, primary WorktreeStore path/name, and workspace terminal paths when a project has exactly one terminal.
- Project custom names are preserved during automatic path sync through the new isNameCustomized field and legacy decode inference.

## UI/UX
- Replaces native .help usage with themed quick tooltips across sidebar, editor, VCS, file tree, tab strip, and command controls.
- Adds a global TooltipOverlay and hides active tooltips before popovers/menu actions so stale tooltips do not linger.
- Makes sidebar resizing much smoother by disabling width animation during drag and only animating collapsed/expanded mode changes.
- Persists sidebar custom width and attached VCS panel width across launches.
- Allows VCS attached panel and file tree panel to be visible together instead of forcing one to close the other.
- Adds a General setting to show or hide top-bar navigation arrows.

## Shortcut and terminal handling
- Supports Tab, Return, Space, and Delete in KeyCombo normalization and SwiftUI KeyEquivalent mapping.
- Makes SwiftUI shortcut binding skip empty key combos instead of crashing on unset shortcuts.
- Fixes terminal Option-as-Alt handling so Option is omitted from consumed_mods when it was stripped for Alt translation.

## Tests
- Added Project decoding, auto-rename, ProjectStore updatePath, and ProjectPathSyncService coverage.
- Added WorktreeStore coverage for primary worktree names following folder names rather than custom project names.
- Added KeyCombo coverage for special key normalization, SwiftUI equivalents, and empty key handling.

## Verification
[...]

PR #311, opened 2026-05-02
✨ feat(split-pane): equal-size split for N panes

dtateks:
## Summary

This PR changes split-pane layout from binary nested sizing to equal sibling sizing, then narrows the default side panels so the terminal workspace starts with more usable space.

The split-pane change is the main feature:

- 2 panes split to `50% / 50%`
- 3 same-direction panes split to `33% / 33% / 33%`
- 4 same-direction panes split to `25% / 25% / 25% / 25%`
- Vertical splits use the same behavior along height
- Mixed directions still create nested branches, so grid-style layouts continue to work

## Problem

Before this PR, `SplitBranch` represented every split as a binary tree:

```swift
first: SplitNode
second: SplitNode
ratio: CGFloat
```

That means splitting an already-split pane created a nested branch instead of adding a third sibling. Example with horizontal panes:

```text
Split once:
[A: 50%][B: 50%]

Split B again:
[A: 50%][B: 25%][C: 25%]
```

This is technically tmux-style behavior, but it produces poor UX for this app because every new pane inherits only the size of the pane it was split from. Users expect all same-direction panes to rebalance across the available workspace.

## New behavior

Same-direction splits now flatten into the existing parent branch and rebalance all siblings:

```text
Split once:
[A: 50%][B: 50%]

Split B again:
[A: 33%][B: 33%][C: 33%]

Split any of A/B/C again:
[A: 25%][B: 25%][C: 25%][D: 25%]
```

Different-direction splits still stay nested, preserving layouts like:

```text
[A: 50%][vertical split: 50%]
         ├── B: 50% height
         └── C: 50% height
```

### Split model

`Muxy/Models/SplitNode.swift`

- Replaced binary fields:
  - `first: SplitNode`
  - `second: SplitNode`
  - `ratio: CGFloat`
- With N-child fields:
  - `children: [SplitNode]`
  - `ratios: [CGFloat]`
- `splitting()` and `splittingWithTab()` now:
  - create a two-child branch when splitting a leaf
  - flatten the new branch into the parent when directions match
  - reset sibling ratios to equal `1 / childCount`
- `removing()` now:
  - removes the target area from N children
  - collapses branches with only one remaining child
  - flattens same-direction child branches
  - rebalances remaining siblings equally
- `areaFrames()` now walks all children with cumulative offsets instead of assuming two frames.

### Split rendering

`Muxy/Views/Workspace/SplitContainer.swift`

- Replaced hard-coded two-child rendering with `ForEach` over `branch.children`.
- Inserts one divider between each adjacent pair.
- Divider drag adjusts only the two adjacent ratios while preserving their combined space.
- Accessibility divider actions still resize with a 5% step.

### Persistence / DTOs

Updated split serialization everywhere the binary split shape was exposed:

- `MuxyShared/WorkspaceDTO.swift`
- `Muxy/Models/WorkspaceSnapshot.swift`
- `Muxy/Extensions/DTOConversions.swift`
- `MuxyMobile/DemoBackend.swift`
- `MuxyMobile/RemoteWorkspaceView.swift`

The serialized shape now uses `children` and `ratios` so desktop, mobile, snapshots, and demo mode all share the same split model.

### Tests

`Tests/MuxyTests/Models/SplitNodeTests.swift`

Added or updated coverage for:

- splitting same direction into 3 equal siblings
- splitting same direction into 4 equal siblings
- preserving nested structure for different-direction splits
- removing panes from 3-child and 4-child branches
- equal frame calculation for 3 children
- custom ratio frame calculation

`Tests/MuxyTests/Models/WorkspaceSnapshotTests.swift`

- Updated split snapshot round-trip for `children` + `ratios`.

## Layout tweak included

This PR also narrows default side panels:

- Sidebar expanded width: `220 → 180`
- File tree default width: `260 → 200`
- Attached VCS default width: `400 → 300`

This gives the terminal workspace more room by default, which pairs with the split-pane UX change.

## Relationship to #308

This PR is intentionally based on `main`, not on #308.

It contains only two commits:

1. `✨ feat(layout): reduce default sidebar and panel widths`
2. `✨ feat(split-pane): equal-size split for N panes instead of binary tree`

The split-pane feature does not depend on the project-path/shortcut changes from #308.

## Verification

Previously run locally on this change set before PR creation:

```text
swift test
Test run with 607 tests in 63 suites passed
```

Also checked PR metadata after creation:

```text
base: main
head: feat/equal-split-pane-clean
commit count: 2
```

### DIFF
--- a/Muxy/Extensions/DTOConversions.swift
+++ b/Muxy/Extensions/DTOConversions.swift
@@ -48,9 +48,8 @@ extension SplitBranch {
         SplitBranchDTO(
             id: id,
             direction: direction == .horizontal ? .horizontal : .vertical,
-            ratio: Double(ratio),
-            first: first.toDTO(),
-            second: second.toDTO()
+            ratios: ratios.map { Double($0) },
+            children: children.map { $0.toDTO() }
         )
     }
 }
--- a/Muxy/Models/SplitNode.swift
+++ b/Muxy/Models/SplitNode.swift
@@ -27,20 +27,17 @@ enum SplitNode: Identifiable {
 final class SplitBranch: Identifiable {
     let id = UUID()
     var direction: SplitDirection
-    var ratio: CGFloat
-    var first: SplitNode
-    var second: SplitNode
+    var children: [SplitNode]
+    var ratios: [CGFloat]
 
     init(
         direction: SplitDirection,
-        ratio: CGFloat = 0.5,
-        first: SplitNode,
-        second: SplitNode
+        children: [SplitNode],
+        ratios: [CGFloat]? = nil
     ) {
         self.direction = direction
-        self.ratio = ratio
-        self.first = first
-        self.second = second
+        self.children = children
+        self.ratios = ratios ?? Array(repeating: 1.0 / CGFloat(children.count), count: children.count)
     }
 }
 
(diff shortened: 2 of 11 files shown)


saeedvaziry:
@dtateks I won't be reading the long AI generated summary. keeping the PR open until you update it and provide some screenshots.

PR #312, opened 2026-05-02
Fix file tree showing git status from parent repo

dtateks:
## Summary

When the file tree is opened at a subfolder inside a git repo (where the subfolder itself has no `.git`), it incorrectly displays git status changes from the entire parent repo. This PR scopes the file tree's git status detection to only include files within the tree's `rootPath`.

## The Bug

**Reproduction:**
1. Open a project whose path is a subfolder of a git repo (e.g. `/repo/sub/` where `/repo/.git/` exists)
2. Open the file tree
3. Modified files in `/repo/` (outside the subfolder) appear as changed in the file tree

**Root cause:** `FileTreeState.loadStatusesSync()` ran `git -C rootPath status --porcelain=v1` and built absolute paths by combining `rootPath` with the git-relative paths returned by git. But git always returns paths relative to the **actual git repository root**, not the directory passed to `-C`. When `rootPath` is a subfolder, this combination produced incorrect paths and pulled in changes from the entire parent repo.

## The Fix

`Muxy/Models/FileTreeState.swift`:
1. **Detect the real git root** via `git rev-parse --show-toplevel` in a new `resolveGitRoot()` helper.
2. **Build absolute paths from the detected git root**, not from the file tree's `rootPath`.
3. **Filter out status entries outside the file tree's scope** — only keep entries whose absolute path starts with `rootPath`.
4. **Canonicalize symlinks** with `realpath()` so `/var` vs `/private/var` (macOS temp dirs) don't break prefix comparisons.

Git review/diff functionality is fully preserved — only the scope detection logic is changed.

## Tests

Added `gitStatusesScopedToTreeRoot` regression test in `FileTreeStateTests` which:
- Creates a real git repo at a temp parent directory
- Opens `FileTreeState` rooted at a subfolder
- Verifies status entries inside the subfolder are present
- Verifies status entries outside the subfolder are excluded
- Verifies nested directories within the subfolder are still marked dirty

All 17 `FileTreeState` tests pass; build is clean.

### DIFF
--- a/Muxy/Models/FileTreeState.swift
+++ b/Muxy/Models/FileTreeState.swift
@@ -351,25 +351,31 @@ final class FileTreeState {
         }
     }
 
-    private struct StatusResult {
+    struct StatusResult {
         let fileStatuses: [String: FileStatus]
         let dirtyDirs: Set<String>
     }
 
-    nonisolated private static func loadStatuses(repoRoot: String) async -> StatusResult {
+    nonisolated static func loadStatuses(repoRoot: String) async -> StatusResult {
         await GitProcessRunner.offMain {
             loadStatusesSync(repoRoot: repoRoot)
         }
     }
 
-    nonisolated private static func loadStatusesSync(repoRoot: String) -> StatusResult {
+    nonisolated static func loadStatusesSync(repoRoot: String) -> StatusResult {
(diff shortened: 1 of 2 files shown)


saeedvaziry:
Another long ai generated summary

PR #313, opened 2026-05-02
Remember file tree and source control panel state per project

dtateks:
## Summary

Adds an opt-in General setting that persists the show/hide state of the attached **File Tree** and **Source Control** side panels independently for each project. When the toggle is on, switching projects restores the last visibility state for that project; when off, behavior is unchanged (shared global state, default hidden).

Also bumps `.tool-versions` to match the currently bottled `swiftformat` 0.61.1 and `swiftlint` 0.63.2 (codebase already passes both).

### Why

Different projects have different workflows — some need the file tree always visible, others rely on source control diff, others want a clean terminal-only view. Today every project share one global toggle, so users have to re-toggle every time they switch.

### Behavior

- **Toggle off (default):** existing global behavior, both panels start hidden, user toggles per session.
- **Toggle on:** opening a project restores its last panel state. Toggling a panel writes the new state back to that project's slot in `UserDefaults`. Panels remain mutually exclusive (opening one closes the other) — same as today.
- Source control panel is only restored when `VCSDisplayMode` is `attached` (consistent with the existing toggle path).

## Changes

- `Muxy/Models/SidebarPanelPreferences.swift` — new store: `rememberPerProject` flag + per-project `[UUID: Bool]` maps for file tree / VCS visibility under `muxy.sidebars.*` keys.
- `Muxy/Views/MainWindow.swift`
  - Added `rememberSidebarsPerProject` `@AppStorage` and `lastRestoredProjectID` state.
  - `toggleAttachedVCSPanel` / `toggleFileTreePanel` now persist visibility for the active project.
  - New `restoreSidebarVisibility(for:)` runs on `onAppear` and on `activeProjectID` change.
  - Extracted into a small `SidebarVisibilityMemorySync` `ViewModifier` (3 small modifiers extracted from the main `body` to avoid a Swift type-checker timeout on the already-large view).
- `Muxy/Views/Settings/GeneralSettingsView.swift` — new "Side Panels" section with toggle: *Remember file tree and source control per project*.
- `.tool-versions` — bump `swiftformat` 0.60.1 → 0.61.1, `swiftlint` 0.57.1 → 0.63.2 (matches Homebrew bottled versions; codebase passes lint+format under both).

## Testing

- [x] `scripts/checks.sh --fix` — Formatting, Linting, macOS Build, Tests pass. iOS build step fails locally only because the runner is missing the iOS 26.4 SDK; unrelated to this change.
- [x] Manual on macOS:
  - Toggle off (default): existing behavior, panels reset across project switches as before.
  - Toggle on:
    - Project A: open file tree → switch to B → open source control → switch back to A. File tree restored on A, source control restored on B.
    - Toggling either panel within a project persists immediately.
    - Source control restoration respects `VCSDisplayMode` (skips when not `attached`).
    - Closing the only active project clears the last-restored marker so re-selecting re-runs restore.

## Screenshots

_To be added by the author._

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-swiftformat 0.60.1
-swiftlint 0.57.1
+swiftformat 0.61.1
+swiftlint 0.63.2
--- a/Muxy/Models/SidebarPanelPreferences.swift
+++ b/Muxy/Models/SidebarPanelPreferences.swift
@@ -1,0 +1,38 @@
+import Foundation
+
+enum SidebarPanelPreferences {
+    static let rememberPerProjectKey = "muxy.sidebars.rememberPerProject"
+    private static let fileTreeStatesKey = "muxy.sidebars.perProject.fileTreeVisible"
+    private static let vcsStatesKey = "muxy.sidebars.perProject.vcsVisible"
+
+    static var rememberPerProject: Bool {
+        get { UserDefaults.standard.bool(forKey: rememberPerProjectKey) }
+        set { UserDefaults.standard.set(newValue, forKey: rememberPerProjectKey) }
+    }
+
+    static func fileTreeVisible(for projectID: UUID) -> Bool {
+        states(forKey: fileTreeStatesKey)[projectID.uuidString] ?? false
+    }
+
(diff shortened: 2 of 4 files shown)


saeedvaziry:
Another long ai generated long summary

PR #346, opened 2026-05-04
Auto-refresh file tree for non-git projects, hide deleted files, add manual refresh

### LINKED ISSUE #345: Default file browser doesn't refresh.
### Description

When project directory structure change (create/edit/delete directories or files) the view doesn't update so the user still see non actual state. No way to refresh. Expected behaviour is that the file tree view is in-sync all the time. 

### Steps to Reproduce

1. Open Muxy
2. Create Project
3. Open files right side bar
4. Create file or dir in the directory
5. Directory structure view is not updated. 

### Muxy Version

0.26.0 (418)

### macOS Version

recent

saeedvaziry:
Fixes #345. The FS watcher used to require a .git directory, so non-git projects never auto-refreshed; it now
  watches any project root. Deleted files are no longer surfaced as synthetic rows — the tree mirrors disk. A manual
  refresh button is added to the header.

### DIFF
--- a/Muxy/Models/FileTreeState.swift
+++ b/Muxy/Models/FileTreeState.swift
@@ -7,7 +7,6 @@ final class FileTreeState {
         case modified
         case added
         case untracked
-        case deleted
         case renamed
         case conflict
     }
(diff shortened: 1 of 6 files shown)


secho-at-revolgy:
Awesome, thanks.

PR #350, opened 2026-05-05
fix: launch external editor commands interactively

sadiksaifi:
## Summary

Fix Terminal Command external editors so commands like `nvim` can resolve PATH setup from interactive shell startup files.

Why: Muxy launched these commands with a login non-interactive shell, so `.zshrc` PATH setup from tools like mise/asdf/nvm was skipped.

Context: https://x.com/thesadiksaifi/status/2051286012501053617

## Changes

- Launch external editor terminal commands interactively.
- Document Terminal Command as login interactive shell behavior.

## Testing

- [x] `swift test`
- [x] `swift test --filter TabArea`
- [x] `scripts/checks.sh --fix`
- [x] `scripts/checks.sh`
- [x] Tested manually on macOS 26.x

## Screenshots

https://github.com/user-attachments/assets/a4e3045f-8f08-4606-9321-4b119c0a0a1b


saeedvaziry:
Thanks @sadiksaifi for the PR. The only issue I have is that this change causes a delay on loading nvim because it first tries to set the env.

sadiksaifi:
> Thanks @sadiksaifi for the PR. The only issue I have is that this change causes a delay on loading nvim because it first tries to set the env.

I initially had the same thought, but then I did some research to understand how other applications address a similar problem. The only reasonable solution that came to mind was this one.

sadiksaifi:
> Thanks @sadiksaifi for the PR. The only issue I have is that this change causes a delay on loading nvim because it first tries to set the env.

I did consider a slightly deeper fix where we resolve the shell env once and reuse it instead of spawning an interactive shell on every editor launch.

That should still handle PATH setup for mise/asdf/nvm while avoiding the repeated startup cost with external command editors.

I avoided it initially to keep this PR scoped, but I can switch to that approach if you think it’s better.

PR #361, opened 2026-05-06
Fix palette ESC key exiting fullscreen instead of closing palette

dunzkoi:
## Summary

Fix palette (Quick Open / Worktree Switcher) ESC key exiting macOS fullscreen instead of closing the palette.

## Changes

Handle cancelOperation: in PaletteSearchField.Coordinator to consume ESC at the palette level

## Testing

## Screenshots

None

### DIFF
--- a/Muxy/Views/Components/PaletteOverlay.swift
+++ b/Muxy/Views/Components/PaletteOverlay.swift
@@ -197,6 +197,10 @@ struct PaletteSearchField: NSViewRepresentable {
             textView _: NSTextView,
             doCommandBy commandSelector: Selector
         ) -> Bool {
+            if commandSelector == #selector(NSResponder.cancelOperation(_:)) {
+                parent.onEscape()
+                return true
+            }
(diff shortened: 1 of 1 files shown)


saeedvaziry:
@dunzkoi this change is totally disables escape button. for example if you open the theme picker on the full screen and hit escape button it closes it but now with this change it doesnt.

This affects other palces like branch selection, worktree, ... where escape closes the popovers but not it doesnt

dunzkoi:
Closing this — couldn't reproduce the original issue (palette ESC exits fullscreen) on current main (build 440). It only reproduces on production build 437, so it looks like something between those builds already addressed it. This PR is no longer needed.

Separate finding worth a new issue: on plain main (without this PR), ESC doesn't close the color theme picker, Switch Worktree popover, or the branch picker. Verified by reverting this PR and clean-rebuilding. Will file a separate issue.

saeedvaziry:
@dunzkoi i could reproduce the issue you mentioned tho

PR #411, opened 2026-05-10
Allow maximizing panes

pedropalmero:
## Summary

Adds a way to temporarily maximize a pane in split workspaces so users can focus on one pane without losing the existing layout.

## Changes

- Add maximize/restore state for split panes.
- Add a pane tab-strip control and Cmd+Opt+Return shortcut.
- Scope tab-number shortcuts while a pane is maximized.

## Testing

## Screenshots

Icon added in tabs

image

Maximized panel

image

### DIFF
--- a/Muxy/Models/AppState.swift
+++ b/Muxy/Models/AppState.swift
@@ -82,6 +82,7 @@ final class AppState {
     var workspaceRoots: [WorktreeKey: SplitNode] = [:]
     var focusedAreaID: [WorktreeKey: UUID] = [:]
     var pendingLayoutApply: PendingLayoutApply?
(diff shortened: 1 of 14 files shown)


saeedvaziry:
@pedropalmero this is a cool feature, however I dont think its a good idea to make the terminal look like island in the maximized mode. just a full size would be enough

pedropalmero:
@saeedvaziry I use this feature it a lot in Zed editor, and I like its appearance. That it looks like floating I think that it makes it easier to distinguish between a regular tab with 100% size and the maximized aspect.

image

saeedvaziry:
I thought a bit and made sense to have it not full size. merged it already thanks

PR #422, opened 2026-05-12
Add session restore feature

IlyaRastorguev:
## Summary

This feature brings abilitty to restore project tabs commands after Muxy relaunch. The commands from project tabs goes throught blacklist, for reason not to run git push or sudo rm.

This feature is not a "tmux-like" behavior, no background tasks running, no client-server, it's more like "continue where you left" in browser.

## Changes

- Add a tabs commands monitoring and saving to local storage (json file for now).
- Add settings section in witch user can enable/disable this feature and also change a commands blacklist
- Add command relaunch mechanism for tabs. This works ONLY when user select project for first time after Muxy relaunch.
- Add per-project closed tabs history. User can now reopen closed tabs with keyboard shortcut (cmd+shift+t by default) and reopened tab opens with previous command.

## Testing


saeedvaziry:
@IlyaRastorguev is this still draft?

IlyaRastorguev:
> @IlyaRastorguev is this still draft?

It's almost done. i still have one bug, it's not "Major" tear, but send to review PR with known bug not in my style 😄.
BTW, @saeedvaziry what do you think about this feature?

saeedvaziry:
Seems a needed feature. I can reserve Cmd+Shift+T for it like browser

IlyaRastorguev:
> Seems a needed feature. I can reserve Cmd+Shift+T for it like browser

This combination already used in tab rename action.

saeedvaziry:
feel free to drop the tab rename shortcut and use for this. we can use shift+option+T for tab rename

IlyaRastorguev:
@saeedvaziry Hi!
I'm want to tell you about one corner case of this feature. If you have multiple tabs with claude, or codex with different chats, after relaunch all of them opens LAST chat, because of restoring method of AI-agents commands. Method modifies command like this:
claude -> claude --continue

I try to make this more smart, but stuck for now. For now i  search the way to get conversation ID of current Muxy tab somehow. If you have some ideas to share with me, i'll be glad to see them.

saeedvaziry:
I think we shouldn't go to that direction at all. I mean ai agent chats.

Muxy is terminal app not ai agent app. So better to rely on terminal and restore the last state only

IlyaRastorguev:
Maybe you'r right. This also make code much simplier. I'm remove this part from PR.

IlyaRastorguev:
@saeedvaziry it's done

Muxy/Services/TerminalLaunchCommand.swift
```diff
@@ -0,0 +1,32 @@
...
+        let flags = interactive ? "-l -i" : "-l"
+        return "\(ShellEscaper.escape(shell)) \(flags) -c '\(script)'"
+    }
+
+    private static var script: String {
+        [
+            "eval \"$\(environmentKey)\"",
+            "muxy_status=$?",
+            "if [ $muxy_status -ne 0 ]",
+            "then exec \"${SHELL:-/bin/zsh}\" -l",
```
IlyaRastorguev: I need to launch the same shell, that user uses, not always zsh. Fix this today.

IlyaRastorguev: done

saeedvaziry:
@IlyaRastorguev I was checking the PR against some performance related guidelines and the problem is this PR writes into disk on every key enter on the terminals. this can defeat the purpose of the app which is memory efficient and lightweight.

Can we for now focus on the goal of the PR only with a smaller amount of changes? Lets just restore on termination. then we can think about the cmd+shift+t

We can get the PR in smaller version of it and merge it and work on the other part

WDYT?

IlyaRastorguev:
@saeedvaziry This is definitely a issue. I will explore this behaviour, can you tell what performance guidelines do you use?

IlyaRastorguev:
I use release build from this branch for a few days and didn't notice any performance changes from last release actually.

saeedvaziry:
> @saeedvaziry This is definitely a issue. I will explore this behaviour, can you tell what performance guidelines do you use?

I use basic performane guidelines such as:

- Memory efficiency (Don't keep memory busy and don't load with heavy stuff)
- efficient and less I/O
- No UI block for tasks that can be async
- ...

Depending on the case I try to check most of these and some more

IlyaRastorguev:
@saeedvaziry I actually can't confirm, that there is a file writing on every keystroke. It's happening only when muxy closes (appDelegate.onTerminate) and on close tab action (normal, force, remote, and with confirmation). Only thing that happening on every keystroke - updating in-memory snapshot of session. I'm optimized saving recently closed tabs logic, file updated two times on one tab close, now this works in right way: one closing - one file updating. Also wrote tests that check that there is no file update on regular terminal input (AppStateTerminalSessionTests.ordinaryTerminalInputTrackingDoesNotSaveTerminalSessions).

saeedvaziry:
@IlyaRastorguev have you tested this on the recent beta?

My honest feedback, I don't find this feature useful. it is causing to boot unwanted tabs even on restore. killing processes closing the tabs and overal it adds heavy memroy usage and load on the startup.

I am thinking to revert it tbh.

saeedvaziry:
btw, Would be nice to have you on Discord server. drop me a msg so we discuss further.

IlyaRastorguev:
@saeedvaziry Of course, I am on a beta update chanel, and use beta builds all time. 

Let me explain, why I start to make this PR.

For now, if you close Muxy and open it again (I am personally do this often), all my open terminal tabs stay open, but it's just plain terminals. And what should I do? I restore all my tabs by myself, switch to first tab, trying to remember what do I have on this tab, then go to the next tab etc, etc. multiply it to five-six projects with several tabs in each we get up to thirty empty terminals that I have to restore manually. So this is why I made this PR.

Close tab on process exit, maybe was bad idea, can agree with you, can easily change this.

About heavy load on startup. I can schedule tab command restore until user select the tab, I think, this also helps to reduce the load time.

I do have discord, and I am will join the Muxy server.

PR #426, opened 2026-05-13
Fix Cmd-click on file paths surfacing 'application can't be opened. -50'

alnaggar-dev:
Ghostty's link auto-detection fires OPEN_URL with a raw absolute path (no file:// scheme), so URL.isFileURL was false and the handler fell through to NSWorkspace.open — macOS then showed 'application can't be opened. -50'.

Route schemeless local paths through AppState.openFile, same as raw-word Cmd-click. Tests added.

### DIFF
--- a/Muxy/Views/Terminal/TerminalPane.swift
+++ b/Muxy/Views/Terminal/TerminalPane.swift
@@ -255,9 +255,8 @@ struct TerminalBridge: NSViewRepresentable {
             Self.resolveFilePath(token, projectPath: projectPath) != nil
         }
         view.onOpenURL = { url in
-            if let projectID, url.isFileURL {
-                let path = url.path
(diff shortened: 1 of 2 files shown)


madciapka:
Independent repro and diagnosis in #427 lines up exactly with this PR — happened to file it a few minutes after this PR went up, didn't see it. Same code pointers (`TerminalPane.swift:257-267`, `GhosttyTerminalNSView.swift:569`), same conclusion that the schemeless-absolute-path OSC 8 case is the trigger, and the suggested fix matches `resolveLocalFilePath` almost line-for-line.

Repro confirmed on Muxy 0.27.0 (471) / macOS 15.7.4 with both `.md` and `.html` files emitted as OSC 8 hyperlinks by `claude` / `codex` sessions. Closing #427 as superseded by this PR. LGTM 👍

PR #428, opened 2026-05-13
Add on-device voice dictation panel

saeedvaziry:
Press Cmd+Shift+I to open a bottom panel that transcribes speech on-device (Apple Speech, no network) and inserts
  the result into the previously focused responder — terminal, editor, or any text field — falling back to the
  clipboard with a toast if the target is gone. Language is configurable in Settings → Recording; status bar gains a
  mic button next to the rich-input toggle.


saeedvaziry:
Superseded by #429 — recreated from main so this PR only contains the voice-dictation commit.

PR #432, opened 2026-05-14
left pane nav panel extending all the way to the top

paulbettner:
Currently the top tab navigation bar extends above the left-hand workspace navigation panel, even though those tabs actually exist within the selected workspace. So I made an adjustment to "fix" this, extending the navigation panel so that it goes to the top of the window.

This is an opinionated UI PR, so please feel free to ignore it... but it makes more sense to me, visually, so I figured I'd submit this PR for your consideration.

Before:

Screenshot 2026-05-09 at 1 02 26 PM

After (this is from my fork):

Screenshot 2026-05-09 at 1 03 06 PM


paulbettner:
Hi @saeedvaziry here's the newly created PR that fixes the access issue. I'll close the other one now.

PR #436, opened 2026-05-14
Improve palette IME input and page navigation

wjchoi87:
## Summary
Stabilizes palette text-field sync during IME composition and adds PageUp/PageDown result navigation.

## Changes
- Avoid overwriting active field-editor text during input.
- Sync committed input before submit.
- Add PageUp/PageDown navigation through search results from the search field.

## Testing

## Recordings
Before: Korean IME composition breaks in the search field.

https://github.com/user-attachments/assets/7f883885-56d0-4680-abe4-f9c984aa5e38

After: Korean IME composition stays stable and PageUp/PageDown navigates results.

https://github.com/user-attachments/assets/1f04e90b-1059-46d1-bdfb-6029684a17fd

### DIFF
--- a/Muxy/Views/Components/FindInFilesOverlay.swift
+++ b/Muxy/Views/Components/FindInFilesOverlay.swift
@@ -48,8 +48,8 @@ struct FindInFilesOverlay: View {
                 onEscape: { onDismiss() },
                 onArrowUp: { moveHighlight(-1) },
                 onArrowDown: { moveHighlight(1) },
-                onPageUp: { moveHighlight(-10) },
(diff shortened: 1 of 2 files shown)


saeedvaziry:
@wjchoi87 this broke the normal english search

wjchoi87:
> @wjchoi87 this broke the normal english search

Let me check the problem

saeedvaziry:
@wjchoi87 ok its not broken. I just noticed a bug that in larger codebases the Cmd+Shift+F causes high cpu usage due to indexing probably

PR #437, opened 2026-05-14
Fix hard-to-grab resize handles

### LINKED ISSUE #375: Resizing borders are thin and not draggable
Inside the app where any resize border like tab pane resize, vcs resize, filetree resize, ... have very thin hnadler that I need to move the cursor exactly on the border to be able to resize.

The expected behaviour is to support wider range for resizing without changing the border width.

wjchoi87:
## Summary

Fixes #375 by expanding resize hit areas without changing the visible border width.

## Changes

- Add a shared resize handle component with a wider invisible hit area.
- Keep resize borders visually 1px wide while expanding the draggable region.
- Highlight resize borders on hover so the active resize area is easier to discover.
- Apply the shared handle to workspace splits, side/bottom panels, and VCS section dividers.

## Testing

## Screenshots

https://github.com/user-attachments/assets/570c86cf-deab-441a-8898-e8dbe6c291f8

### DIFF
--- a/Muxy/Theme/UIMetrics.swift
+++ b/Muxy/Theme/UIMetrics.swift
@@ -37,7 +37,7 @@ enum UIMetrics {
     static var controlSmall: CGFloat { scaled(20) }
     static var controlMedium: CGFloat { scaled(24) }
     static var controlLarge: CGFloat { scaled(32) }
-    static var resizeHandleHitArea: CGFloat { scaled(12) }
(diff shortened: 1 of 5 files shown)


saeedvaziry:
image

there is this empty space happening which shouldn;t be

wjchoi87:
> image
> there is this empty space happening which shouldn;t be

I didn't take a look, I'll work again.

wjchoi87:
Fixed the issue where blank spaces were appearing.
Please review this.

PR #460, opened 2026-05-18
feat: add project groups to organize projects in sidebar

ejklock:
Adds a project groups layer to the sidebar so projects can be organized into named, collapsible sections.

## What's included

- **Model + persistence**: `ProjectGroup` struct with id, name, sortOrder, isExpanded, and projectIDs. `ProjectGroupStore` handles CRUD, reorder, and migration (first launch creates a Default group with all existing projects).
- **Sidebar rendering**: Groups render as collapsible sections via `ProjectGroupSection`. Collapsed sidebar shows the group initial letter as a badge. Each group is wrapped in a rounded border for visual separation.
- **Drag-and-drop**: Group-aware reorder — drag within the same group reorders, drag between groups moves the project, drag from ungrouped into a group assigns it. Group headers are also valid drop targets for collapsed groups.
- **Context menus**: Group header has Add/Rename/Delete. Project rows have Move to Group (including ungrouped projects).
- **Wiring**: `ProjectStore.onProjectAdded` auto-assigns new projects to the first group. `onProjectRemoved` cleans up group references. Callbacks set in `MuxyApp.onAppear`.

## Test coverage

- `ProjectGroupStoreTests` (18 tests): CRUD, reorder, migration (empty+projects, empty+empty, existing groups), cross-group move, duplicate guard.
- `SidebarGroupTests` (11 tests): group section rendering, context menu filtering, sequential shortcut indexing across groups.

🤖 Co-authored with Claude Code


saeedvaziry:
image

Its a good idea but hoenstly I don't like the UI and UX of this feature

ejklock:
> image
> 
> 
> 
> Its a good idea but hoenstly I don't like the UI and UX of this feature

I confess that I'm not too good with ux 😅. But do you have any thoughts about how need to be this feature? I'm open and happy to help

Tks

IlyaRastorguev:
I think, that grouping by default is bad idea, more convenient way is to give user an ability to create project groups, but don't do this for him (auto default group creation). Also i want to notice that this will be third collapsing level:
Group -> project -> worktree
so this maybe difficult to operate with for some people
Simplicity is the way 🤗

IlyaRastorguev:
Maybe, you can make something like safari's tabs grouping, or many others browsers also have something like workspaces. 
image
image
This is more clean way i think.

@saeedvaziry what do you think?

saeedvaziry:
I would appreceate a prototype to see the idea in practice @IlyaRastorguev

saeedvaziry:
There is also this idea that we dedicate the sidebar to the selected project/workspace and show make sections like worktrees, commands, and in the future notes, tasks, ...

ejklock:
> Maybe, you can make something like safari's tabs grouping, or many others browsers also have something like workspaces. image image This is more clean way i think.
> 
> @saeedvaziry what do you think?

i liked this idea too... i'm not a safari user and dint saw this before... tks

ejklock:
@saeedvaziry @IlyaRastorguev what about this?

image
image
image
image

image

saeedvaziry:
This looks a better alternative. But I'd like to take over this PR and add more stuff if you don't mind

ejklock:
> This looks a better alternative. But I'd like to take over this PR and add more stuff if you don't mind

Sure! Go ahead

iodic:
Guys, throwing in another option:

muxy sidebar groups

The idea here is that the groups don't collapse so we maintain visibility while making sure the hierarchy is clear.

An alternative where we could collapse is this:

muxy sidebar groups-alt

Disclaimer: these are rough (alignment and overlapping issues…) but I hope that they illustrate the approach well enough.

saeedvaziry:
@iodic they look cool but the idea is to make it more focused and yet easy to switch between projects and make room for more features next to worktrees like commands, tasks, ...

iodic:
@saeedvaziry, thanks for the feedback!

If the sidebar is going to become 'busier' per project, with workspaces/tasks/commands/... sections, I would suggest using workspaces for grouping projects. Similar to how the Arc browser uses workspaces:

muxy-sidebar-workspaces

I know this would make things more complicated but the separation will be clearer: one sidebar pane per project group/client…

saeedvaziry:
@iodic the last screenshot looks better

IlyaRastorguev:
> @saeedvaziry, thanks for the feedback!
> 
> If the sidebar is going to become 'busier' per project, with workspaces/tasks/commands/... sections, I would suggest using workspaces for grouping projects. Similar to how the Arc browser uses workspaces:
> 
> muxy-sidebar-workspaces
> I know this would make things more complicated but the separation will be clearer: one sidebar pane per project group/client…

Looking good, yes, but you need to specify workspace icon for every workspace, and this is the point of cognitive load starts growth. First you must find the icon, that associate with projects in this workspace, this may be a not easy task (my work projects usually do not have any icons, because i work as backend dev). Second, you always must remember, witch icon for witch workspace you set. I think, that we can take last @ejklock variant as a base idea and make it good looking. Or just create a selector from workspace name on the top of sidebar.
image
         ^ ^ ^
          here
In collapsed sidebar we can just show button with workspaces icon, maybe this box icon (box with projects 😄)
image

saeedvaziry:
> I think, that we can take last @ejklock variant as a base idea and make it good looking

This is I am sure we are takling for this PR as it is simple and solid.

For the commands/worktrees/... lets think after merging this so we can compare the prototypes

PR #464, opened 2026-05-19
fix(process): add timeouts for Git and helper subprocesses

swaynehales:
## Description

This came up while looking into app instability...

Currently a hung helper subprocess can leave the app in indefinite wait (and the user lol).

This adds bounded subprocess handling for:

- Git commands, with timeout budgets for metadata, status/diff, and network operations.
- File search and text search helpers.
- AI usage token reads.
- Mobile `lsof` discovery.
- File-tree ignore checks.

A shared timeout watcher sends best-effort process-group termination first, then force-kills if the process is still running.

## Related Issue

N/A

## Potential Risk & Impact

Very slow helper commands may now be terminated instead of running indefinitely. Git network operations have the longest timeout budget, but unusually slow networks may need a retry.

## How Has This Been Tested?

- `swift test --filter 'GitProcessRunnerCredentialHelperTests|TextSearchServiceTests'`
- `PATH="$PWD/.build/tools/bin:$PATH" scripts/checks.sh --fix`

### DIFF
--- a/Muxy/Services/AIUsageTokenReader.swift
+++ b/Muxy/Services/AIUsageTokenReader.swift
@@ -56,9 +56,11 @@ enum AIUsageTokenReader {
             return nil
         }
 
+        let watcher = ProcessTimeoutWatcher.install(on: process, timeout: 5)
         let outputData = stdout.fileHandleForReading.readDataToEndOfFile()
         _ = stderr.fileHandleForReading.readDataToEndOfFile()
         process.waitUntilExit()
+        watcher.cancel()
 
         guard process.terminationStatus == 0,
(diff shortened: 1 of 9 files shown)


saeedvaziry:
@laywens I believe we could have alternative solutions maybe. for example I've seen a lot of real work repos that are heavy and adding network delay on top can take longer push and pulls.

An alternative would be to somehow cancel these operations rather than bound to timeouts

swaynehales:
Closing as part of fork branch consolidation. I am retiring this stale upstream PR branch to reduce maintenance debt; the branch tip has been archived locally before deletion.

PR #465, opened 2026-05-19
build(release): harden signed release packaging

swaynehales:
## Description

Came up while testing out local signed builds, seemed worth bringing over...

This updates the release script so that:

- Auto-selected signing identities are passed to `codesign` by certificate hash.
- The selected identity is still printed with a readable label.
- The `create-dmg` install hint points to pnpm.
- The target DMG is removed before packaging.
- The script fails if the expected DMG is not produced.

## Related Issue

N/A

## Potential Risk & Impact

Release packaging will now fail earlier when signing or DMG creation isn't set up, rather than producing a stale or mis-signed artifact.

## How Has This Been Tested?

- `bash -n scripts/build-release.sh`
- `swift build -c release`
- `PATH="$PWD/.build/tools/bin:$PATH" scripts/checks.sh --fix`
- `scripts/build-release.sh --arch arm64 --version 0.0.0-beta.516 --sign`
- `codesign --verify --deep --strict --verbose=2 build/Muxy.app`
- `codesign --verify --strict --verbose=2 build/Muxy-0.0.0-beta.516-arm64.dmg`

### DIFF
--- a/scripts/build-release.sh
+++ b/scripts/build-release.sh
@@ -61,16 +61,36 @@ if [[ "$AUTO_SIGN" == "true" && -n "$SIGN_IDENTITY" ]]; then
     exit 1
 fi
 
+find_signing_identity() {
+    local certificate_prefix="$1"
+
+    security find-identity -v -p codesigning | awk -v prefix="$certificate_prefix" '
+        index($0, "\"" prefix) && hash == "" {
+            hash = $2
+            name = $0
+            sub(/^[^"]*"/, "", name)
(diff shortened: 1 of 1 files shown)


saeedvaziry:
What is the reason switching to `pnpm`?

swaynehales:
Closing as part of fork branch consolidation. I am retiring this stale upstream PR branch to reduce maintenance debt; the branch tip has been archived locally before deletion.

PR #441, opened 2026-05-15
Add custom project picker, replaces NSOpenPanel

sadiksaifi:
## Summary

Adds a custom in-app project picker so opening/adding projects is faster and doesn’t always kick you into Finder.

Closes https://github.com/muxy-app/muxy/issues/385

## Changes

- Added the custom project picker for opening, adding, and creating project folders.
- Wired the sidebar add button and open-project shortcut to use the picker by default.
- Kept Finder available from the picker menu and as a Settings option.
- Added path typing, folder browsing, ghost text, and `Tab` autocomplete.
- Added keyboard shortcuts for navigate, autocomplete, open, add/create, go back, and close.
- Added footer shortcut hints, including `Tab` for autocomplete.
- Added picker settings for `Custom` vs `Finder`.
- Added configurable default picker location with reset and invalid-location warnings.
- Added project-open handling for existing projects, new folders, duplicates, and failed paths.
- Split picker behavior into smaller testable services for state, commands, paths, routing, and settings focus.
- Added tests for picker routing, path handling, project opening/creation, settings, session behavior, and shortcuts.

## Testing

- [X] `scripts/checks.sh` passes
- [X] Tested manually on macOS

## Screenshots

https://github.com/user-attachments/assets/830e4327-89b5-4f4c-ae38-8bc5f14c99aa

PR #482, opened 2026-05-20
Add new OmniBox Overlay

IlyaRastorguev:
## Summary

This PR brings new OmniBox overlay for searching projects, worktrees, opened tabs, recentry closed tabs and also a custom commands.
The search is scope based, first level - projects, after selecting project scope changes to worktrees, then activates three scope regime - tabs/history/commands, but user can change that with hotkeys, to activate only one scope.
Also user can create a new tab with specific command.

Hotkeys:
Shift+Cmd+o - shows Omnibox
Tab/Shift+Tab/arrow keys - select next/previous list item
Opt+Cmd+p - change scope to Projects
Opt+Cmd+w - change scope to Worktrees
Opt+Cmd+t - change scope to Tabs (works only when Project and worktree already selected)
Opt+Cmd+h - change scope to recent tabs history (works only when Project and worktree already selected)
Opt+Cmd+c - change scope to custom commands (works only when Project and worktree already selected)
Esc - close overlay
Enter - select

This PR adds common, shared Overlay UI component (OverlayPanel.swift), based on [this PR](https://github.com/muxy-app/muxy/pull/441) code, and replaces old OpenerOverlay.

## Testing

## Screenshots

image


saeedvaziry:
@IlyaRastorguev this looks great but I would be more a fan of simplicity here.

Right now with this PR, if I wanna switch workspaces I have to first hit one backspace then select worktree and then select a tab of a worktree and then finally be able to switch the worktree.

Scopping is good but only for visibility I'd say.

We could show all the results and give user only a text box to search because its way more convenient. however user can toggle what they are looking for usually in the filter box 👇

image

and more importantly, these filters can be persisted for the next run. so some might just wanna use the OmniBox for switching between worktrees, or some wanna switch between tabs! and some both but open to navigate with keyboard or type more for more accurate result.

WDYT?

IlyaRastorguev:
I'm thinking about this to, but when I trying my first build with no scoping, I had a dozen results in list with the same title, my tabs at most are the same in every project and every worktree - neovim, claude, lazygit, ets so it was difficult to select what I need with a first try 😃. Maybe I can add a shortcuts, that opens omnibox with active scope?

Cmd+shift+w for worktrees,
Cmd+shift+p for projects.

And for this mode I can make opening project or worktree immediately, without selecting all next things.

saeedvaziry:
I think making the filters persistent would solve that issue. You filter once and thats it

IlyaRastorguev:
I made update to PR, that brings separate modes to OmniBox, each of them can be open with different shortcuts:
opt+cmd+p - projects
opt+cmd+w - workspaces
opt+cmd+o - tabs
opt+cmd+h - tabs history
opt+cmd+c - custom commands

PR #486, opened 2026-05-20
[feat] add Option key behavior setting

### LINKED ISSUE #485: [bug] codex thinking level shortcut opt+, doesn't work
### Description

Codex thinking level shortcuts `opt+,` and `opt+.` don't work as intended – instead printing regular `≤ ≥` symbols.
In Ghostty I've used this setting in config `macos-option-as-alt = true` but can't find analog in Muxy settings window.

### Steps to Reproduce

1. Open Muxy, start codex
2. Press opt+, and opt+.
3. Expect thinking level to change. Instead ≤≥ appears.

### Muxy Version

0.27.0

### macOS Version

26.5

dergachoff:
Problem

Codex supports alt+, & alt+. for changing thinking levels. But by default Ghostty (and Muxy) treat opt and alt as different keys, so the shortcut doesn't work and produces ≤≥ instead.

Muxy config file `~/Library/Application Support/Muxy/ghostty.conf` supports `macos-option-as-alt = true` but there's no dedicated setting field in UI.

Fix

In this PR I've added a dropdown with all options supported by Ghostty. `Settings > General > Terminal > Option key behavior`. Automatic preserves default by leaving config key unset. The other options write `true`, `false`, `left`, or `right` to Muxy's managed `ghostty.conf`.

Verification
Roses are red, test suite is green
- scripts/checks.sh --fix
- Manually verified Settings > General > Terminal updates ghostty.conf and the selected dropdown value stays visible

Closes #485


saeedvaziry:
@dergachoff have you tried this with ghostty config? top Muxy menu on mac topbar > Open Configuration...

dergachoff:
@saeedvaziry yes, I've found ghostty settings text file later (not through menu btw, it looks obvious in hindsight, but i was looking in settings, then found while manually checking install folders). 

But since Muxy looks and feels more like proper desktop app with gui and rich settings menu, I though that it's reasonable to show settings to user in a proper place. 

And it also feels that for end user this whole notion of ghostty vs muxy settings is quite confusing – "i use muxy, i need to find and adjust settings, so i go to settings" – ghostty is a backend is just implementation specifics

saeedvaziry:
I can live with this but what I see is, if I add this one into Muxy settings, I am opening the path many PRs adding other things from Ghostty. Maybe is a good idea to add it but I will park this until I refactor the settings window to organize it better and support search so we can add more cases.

dergachoff:
Another example: in ghostty i changed theme in text file, in muxy - in settings gui. So then I expect to be able to change anything adjustable in settings window. And not think "what's ghostty concern? what's muxy concern?"

This patch is codex-specific and easily searchable by frustrated codex users, I'm intentionally keeping it narrow. 

But honestly I think having settings parity in GUI is a proper way (at least to my taste).

dergachoff:
@saeedvaziry makes sense. Don't know if it should be separated into ghostty zone and muxy zone, feels that for end user it should be just a single well-organized settings window covering all concerns in their own sections without leaking backend implementation details.

saeedvaziry:
That theme is not for ghostty. We have another layer specifically for the themes applying to libghostty from the app

saeedvaziry:
@dergachoff anyway, I am not against the PR. I am saying we should do it proper rather than one by one adding options

dergachoff:
@saeedvaziry you're the author, I definitely don't object :) Thinking about bigger picture of how settings should be organized makes sense.

saeedvaziry:
Closing this until in case we add full ghosty config support

PR #494, opened 2026-05-21
feat(editor): add HTML file preview

nguyenvanduocit:
## Need
Preview rendered HTML files inside Muxy instead of reading raw source or bouncing to a browser.

## Problem
HTML files only opened as plain source in the code editor — seeing the actual page meant opening it in an external browser, and the existing Code/Preview/Split toggle was wired to Markdown only.

## Solution
Add an HTML preview mode. `.html`/`.htm` files open with the same Code/Preview/Split toggle, defaulting to Preview, and the mode picker is generalized so Markdown and HTML share it. Preview renders in a `WKWebView` via `loadFileURL` with read access to the file's directory, so relative assets, stylesheets, and scripts resolve; JavaScript is enabled and ⌘R reloads. Covered by `EditorTabStateTests`.

image


saeedvaziry (approved):
Approved. Checks passed and I found no required changes.

PR #496, opened 2026-05-21
feat(file-tree): add hide-ignored-files toggle

nguyenvanduocit:
## Need
Declutter the file tree by hiding files that aren't part of the work — git-ignored build output, dotfiles, and dependency noise.

## Problem
Git-ignored entries were always shown, only dimmed to `0.45` opacity, with no way to hide them. Tracked dotfiles (`.github`, `.omc`) and `node_modules` cluttered the tree, and the only filter was "Show only changes".

## Solution
Add a header toggle (eye icon) next to "Show only changes" that hides git-ignored entries, dotfiles, and a built-in noise list (`node_modules`, common lockfiles). Off by default, persisted globally in `UserDefaults`. It composes with the changes filter; the file open in the editor stays visible even when it matches a rule. Covered by `FileTreeStateTests` — classification, filtering, persistence, and the reveal exemption.


nguyenvanduocit:
I want to bring all my need to this app, at this point, it is so personal, I think we should develop plugin system, nowerday, AI can do alot, personal app is a must, but we can not fork for every one, no way to sync with upstream. Maybe plugin is the way.

saeedvaziry:
@nguyenvanduocit already thought about the plugin idea. Its a bit tricky and risky. tho I am considering a limited API for the plugins but its an open discussion yet. Feel free to open a thread in Discord.

nguyenvanduocit:
I dont know if this MR is personal need or not, but I very love this feature, nowerday config folders, files is everywhere, which we often no need to see.

saeedvaziry:
anyway, this feature is standard. will review and merge

PR #505, opened 2026-05-22
Fix Git panel flicker on non-git projects

### LINKED ISSUE #506: Git panel flickers continuously on non-git projects
## Description
Opening the attached Git panel (VCS sidebar) on a folder that is not a git repository makes the panel flicker continuously — it rapidly alternates between the "not a git repository" message and a full but empty git UI (commit box, Commit/Pull/Push buttons, Changes / History / Pull Requests sections).

## Steps to reproduce
1. Add a project/folder that is not a git repository.
2. Open the attached Git panel for it.
3. With any normal filesystem activity in the folder (a terminal running in it, an editor, or macOS writing `.DS_Store`), the panel flickers several times per second.

The first open shows a clean spinner → error (one transition). The continuous flicker starts once the file-system watcher begins firing.

## Expected
The panel shows a stable "not a git repository" state.

## Actual
The panel flickers between the error message and a blank git UI.

## Root cause
`VCSTabState.performRefresh()` clears `errorMessage = nil` synchronously, but only restores it later in the async `loadFilesTask` catch block after `git rev-parse` fails. During that gap `VCSTabView.content` evaluates `files.isEmpty && errorMessage != nil` as false and renders the full git UI. `FileSystemWatcher` drives this on every filesystem event because its filter only suppresses `.git/`-internal events, which a non-git folder has none of.

## Environment
- Muxy, `main`.
- macOS.

Fix: #505

nguyenvanduocit:
## Summary
Opening the Git panel on a non-git folder made it flicker between the "not a git repository" message and a blank git UI. `performRefresh` cleared `errorMessage` synchronously but only restored it after the async git lookup failed, so every file-system event briefly rendered the full git UI. `errorMessage` is now cleared only when a refresh completes successfully.

Closes #506

## Testing
`scripts/checks.sh` passes — formatting, linting, build, and the full test suite. Verified manually on a non-git folder: the Git panel stays on the error message instead of flickering.

## Screenshots / Recording
Before: the panel flickers between the error message and an empty git UI on a non-git folder.
After: the panel stays on the "not a git repository" message.

### DIFF
--- a/Muxy/Models/VCSTabState.swift
+++ b/Muxy/Models/VCSTabState.swift
@@ -308,7 +308,6 @@ final class VCSTabState {
         }
         isRefreshing = true
         pendingRefresh = false
-        errorMessage = nil
 
         let refreshSignpost = GitSignpost.begin("performRefresh", incremental ? "incremental" : "full")
 
@@ -413,6 +412,9 @@ final class VCSTabState {
                 let listChanged = files.map(\.path) != newFiles.map(\.path) || !changedPaths.isEmpty
                 if listChanged {
                     files = newFiles
+                }
+                if errorMessage != nil {
+                    errorMessage = nil
                 }
                 isLoadingFiles = false
                 hasCompletedInitialLoad = true


saeedvaziry (approved):
Approved. The change keeps the existing non-git error visible until a successful refresh, which fixes the flicker without broad state changes. Local checks: formatting/lint/build passed; initial full suite had a transient test failure, rerun of swift test passed all 924 tests.

PR #529, opened 2026-05-23
feat: add Pi AI provider notification support

maplezzk:
### Summary

Add Pi coding agent support to Muxy's AI Providers notification system.

### Changes

- **PiProvider.swift**: New `AIProviderIntegration` provider that installs a TypeScript extension into Pi's `~/.pi/agent/extensions/` directory
- **muxy-pi-extension.ts**: Pi extension subscribing to `agent_end` event, sending notifications to Muxy's Unix socket
- **pi.svg**: Provider icon for notification panel
- **AIProviderIntegration.swift**: Registered PiProvider in the registry
- **PiProviderTests.swift**: 11 unit tests, all passing

### Architecture

Pi uses TypeScript extensions (not shell hooks like Claude Code/Droid), so PiProvider:
1. Locates the bundled `.ts` extension file via `Bundle.appResources.url(forResource:withExtension:)`
2. Copies it to `~/.pi/agent/extensions/muxy-notify.ts` (auto-discovered by Pi on next startup or `/reload`)
3. The extension subscribes to `agent_end` and sends notifications through the Muxy socket

### Notes

- Bypasses the hardcoded `.sh` extension in `installAll/forceInstall` by directly accessing the bundle resource


maxschulmeister:
The extension is actually not being copied automatically in my case.
I updated rather than installed the new version, so that might be the issue. 
Couldn't test it yet.

maxschulmeister:
@maplezzk I filed a PR that fixes my comment here.

https://github.com/muxy-app/muxy/pull/566

This is actually the cleaner approach imo, as we don't modify the pi instance outside of Muxy.

PR #538, opened 2026-05-24
Add fullscreen shortcut

leaxoy:
Support toggle fullscreen use shortcut `cmd+ctrl+f`


saeedvaziry:
What happened to the text finder which was using that shortcut?

saeedvaziry:
added in the latest beta

PR #548, opened 2026-05-25
feat: add 'Open with Default App' in sidebar and terminal context menus

maplezzk:
Adds the ability to open files with the system default application from two entry points:

1. **Sidebar file tree right-click menu** — new "Open with Default App" menu item
2. **Terminal right-click menu** — when right-clicking on a file path in the terminal, "Open with Default App" appears at the top of the context menu

Both use `NSWorkspace.shared.open` to launch the file with the macOS default handler (e.g. .dmg with DiskImageMounter, .pdf with Preview, etc.).

Relative paths in the terminal context menu are resolved against the pane's current working directory.


maplezzk:
img_v3_02121_9b1118ee-8313-4d24-b235-384084a2c38g
image

PR #564, opened 2026-05-26
fix: resolve hook script not found when stored in bundle scripts/ subdirectory

maplezzk:
Bundle.url(forResource:withExtension:) does not search subdirectories, which caused hook scripts under `scripts/` in Muxy_Muxy.bundle to be unresolvable.

**Root cause:** All hook scripts (muxy-claude-hook.sh, muxy-pi-extension.ts, etc.) are stored in the `scripts/` subdirectory, but `url(forResource:withExtension:)` only searches the bundle root. This affected **every provider** — none of them could install hooks.

**Changes:**
- Make `findBundledScript` also try the `scripts/` subdirectory as fallback — fixes the core issue for all providers
- Fix `PiProvider` resource URL separately — it ignores the `hookScriptPath` parameter and reloads the file from bundle itself, so it needed an additional fix
- Create `settings.json` if it doesn't exist during extension registration (previously silently skipped)
- Add unit tests for bundled script resource lookup (root file, scripts/ subdirectory, missing file, precedence)


maplezzk:
## Testing verification

Both notification plugins have been tested and confirmed working:

### Pi
- ✅ Hook script (muxy-pi-extension.ts) installed to `~/.pi/agent/extensions/muxy-notify.ts`
- ✅ Extension registered in `~/.pi/agent/settings.json`
- ✅ Uninstall cleans up both file and settings
- ✅ Reinstall works correctly

### Claude Code
- ✅ Notification and Stop hooks registered in `~/.claude/settings.json`
- ✅ Uninstall removes all hooks
- ✅ Reinstall works correctly

### Unit tests
- ✅ 4 tests added covering root file, scripts/ subdirectory, missing file, and precedence scenarios

Issue #558, opened 2026-05-26
Generting commit message with Pi fails


maxschulmeister:
### Description

"Provider returned an empty response."

test with custom model and default model.

### Steps to Reproduce

1. Open Muxy
2. Ensure Settings set Pi and default or custom model as AI Assistant
3. Make file changes
4. Open Source Control Panel via Muxy
5. Click AI Commit generation icon
6. Observe Error

### Muxy Version

0.29.0-beta.582 (582)

### macOS Version

26

maxschulmeister:
On a sidenote, the action is saved as a session in Pi which shouldnt be the case and is avoidable.

maplezzk:
OK, let me take a look. I'll fix it as soon as possible

maxschulmeister:
Investigated this on my side. Muxy’s integration looks correct — the error "Provider returned an empty response." comes from
 AIAssistantRunner when the subprocess exits 0 but stdout is empty after trimming.

 Muxy runs the provider like:

 ```bash
   zsh -l -i -c '<command>'
   # prompt → stdin, response ← stdout
 ```

 That matches pi’s documented print-mode usage (cat … | pi -p "…").

 ### Root cause (upstream in pi)

 With pi 0.75.5, text print mode (pi -p) returns empty stdout when the prompt arrives via piped stdin, even though the model responds:

 ```bash
   echo "say ok only" | pi -p --no-session --provider openai-codex --model gpt-5.4-mini
   # exit 0, stdout: empty
 ```

 Same with Muxy’s built-in Pi command (pi -p + optional --model).

 pi -p "say ok only" does work in an interactive terminal because the prompt is a CLI argument and stdin is a TTY — but Muxy pipes the prompt on stdin, so that path doesn’t apply.

 pi --mode json does stream output during the run; only the final text extraction in plain -p mode is lost because the process exits before session.prompt() fully settles.

[ I’ve filed a bug upstream in pi.
](https://github.com/earendil-works/pi/issues/5031)

 ### Workaround (Custom Command in Settings → AI)

 Use JSON mode and extract the last assistant text block:

 ```bash
[...]

PR #565, opened 2026-05-26
fix: add --no-session flag to Pi assistant invocation

maplezzk:
Related to #558

Add `--no-session` flag to prevent Pi from saving a session when Muxy calls it for commit message generation via stdin pipe.


saeedvaziry:
Closing as this feature has been dropped from Muxy

PR #566, opened 2026-05-26
Install Pi notifications via app-local pi wrapper

maxschulmeister:
Continues #529 with an app-local install path that avoids mutating `~/.pi`.

## Summary
- Install Pi notification extension and a Muxy-owned `pi` wrapper under Application Support instead of mutating `~/.pi`
- Prepend the Muxy `bin` directory to terminal `PATH` so Pi sessions use the wrapper automatically
- Migrate away the legacy global extension install and bump SwiftLint to 0.63.2

## Test plan
- [ ] Run `scripts/checks.sh`
- [ ] Install Pi provider from Muxy settings and confirm files land in `~/Library/Application Support/Muxy/bin`
- [ ] Open a Pi terminal pane and verify `which pi` resolves to the Muxy wrapper
- [ ] Trigger a Pi notification and confirm it appears in Muxy with click-to-navigate

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 swiftformat 0.60.1
-swiftlint 0.57.1
+swiftlint 0.63.2
 xcode 16.4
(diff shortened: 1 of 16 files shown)


saeedvaziry:
Closing as stale and most of these files already dropped. I'd appreceate if you merge the 2 Pi related PRs into one and send it over.

Issue #563, opened 2026-05-26
Confirm when deleting a project or a worktree.


loveforgeter:
### Problem

当我在右键点击项目进行操作的时候，有可能会误点击到delete选项，导致整个project或者worktree-branch被删除。期望能加入确认对话框。
----
When I right-click a project to perform operations, I may accidentally click the delete option, resulting in the deletion of the entire project or worktree-branch. It would be appreciated if a confirmation dialog box could be added.

### Proposed Solution

删除分枝或者项目时二次确认。
----
Add a secondary confirmation prompt when deleting branches or projects.

PR #582, opened 2026-05-27
fix(sidebar): confirm removals

QuentinHsu:
## Summary

- Adds confirmation before removing projects or worktrees from the sidebar.
- Reduces accidental destructive actions, especially when deleting worktrees from disk.

#563

## Changes

- Adds a project removal confirmation alert in the sidebar.
- Uses the same SwiftUI alert style for worktree removal from both sidebar entry points.
- Makes Remove the default action so pressing Enter confirms the dialog.
- Keeps warning copy for clean and dirty worktree removal cases covered by tests.

## Testing

- [x] `scripts/checks.sh --fix` passes
- [x] `swift test --filter WorktreeRemovalConfirmationTests` passes

## Screenshots

CleanShot 2026-05-27 at 16 00 42
CleanShot 2026-05-28 at 12 08 49@2x


saeedvaziry:
@QuentinHsu from the screenshot seems different dialogs for each.

can we have same dialog first screenshot and with confirm button focused already so hitting enter on keyboard on the dialog open will confirm it?

QuentinHsu:
> @QuentinHsu from the screenshot seems different dialogs for each.
> 
> can we have same dialog first screenshot and with confirm button focused already so hitting enter on keyboard on the dialog open will confirm it?

Done. Both removal confirmations now use the same dialog style as the first screenshot, and the confirm button is the default action so pressing Enter confirms it when the dialog opens.

Also ran `scripts/checks.sh --fix` successfully.

PR #605, opened 2026-05-29
feat(sidebar): auto-refresh worktrees on external changes

### LINKED ISSUE #604: [Feature Request] Auto-refresh project worktree list (no manual Refresh needed)
### Problem

When worktrees are added or removed outside Muxy (e.g. running git worktree add / git worktree remove in a terminal, or via another tool), the sidebar worktree list doesn't update. You have to manually open the project's context menu and click Refresh Worktrees to see the change.

When a project's worktrees are added or removed, the sidebar refreshes automatically, with no manual action.

### Proposed Solution

Watch each project's .git directory via FSEvents and trigger the existing WorktreeRefreshHelper.refresh only when a worktree admin directory (.git/worktrees/<name>) is created/removed/renamed:

- Watch .git (always exists) rather than .git/worktrees (which doesn't exist until the first worktree is created).
- Filter: path contains /worktrees/ + is a directory + Created/Removed/Renamed → commits inside a linked worktree (file-only changes) won't trigger a refresh.
- 0.3s debounce, reusing VCSWorktreeAutoRefresher's existing coalescing; keep manual Refresh as a fallback.

### Additional Context

_No response_

passion729:
## Problem
When worktrees are added/removed outside Muxy (terminal `git worktree add/remove`), or a branch is switched/renamed, the sidebar doesn't update — you have to manually click "Refresh Worktrees". Muxy's own operations update the store; the gap is only externally-driven changes that go undetected.

## Changes
- **New `GitWorktreesWatcher`**: an FSEvents watcher on each project's `.git` directory (always present, which avoids the chicken-and-egg of `.git/worktrees` not existing until the first worktree). On a relevant change it reuses the existing `WorktreeRefreshHelper.refresh`.
- **Precise filtering** so it only refreshes on changes that matter:
  - Directory create/remove/rename of `.git/worktrees/<name>` → catches worktree add/remove.
  - `HEAD` symref writes (primary `.git/HEAD` and linked `.git/worktrees/<name>/HEAD`) → catches branch switch/rename.
  - Excludes `logs/HEAD` (reflog, written on every commit) and `.lock` files, so **commits do not trigger a refresh** (avoids noise).
- **Wired into `VCSWorktreeAutoRefresher`**: uses `withObservationTracking` to create/tear down watchers as `projectStore.projects` changes, and reuses its existing in-flight/pending coalescing. 0.3s debounce.
- **Fix stale name on branch rename**: when a worktree's `name` was auto-derived from its branch (`name == old branch`), it now follows the renamed branch on refresh; user-customized names (`name != branch`) are preserved.

## Notes
- Pure background behavior, no UI changes; manual "Refresh Worktrees" stays as a fallback.
- Negligible idle cost: the callback only does string/flag checks on a background queue, and `git worktree list` runs only on real structural/branch changes.
- FSEvents can be unreliable on some network volumes; the manual Refresh remains available there.

## Testing
- `swift build` ✅
- `swift test --filter GitWorktreesWatcher` (11) and `--filter WorktreeStore` (12) all green, including live FSEvents triggers (dir creation / HEAD write) and the rename name-sync case.
- swiftformat / swiftlint --strict clean on the changed files.

Closes #604


saeedvaziry:
We dropped `rg` and cant accept another binary to the core. consider making an extension for this

passion729:
@saeedvaziry Good point. The bundled `rg` slipped into this PR but it is not required for the worktree watcher change, so I’ll remove it from core.

For this PR, the core change is only the FSEvents-based `.git` watcher that refreshes worktrees when external git worktree/branch changes happen. That does not need any new binary dependency. If we still want richer search later, I’ll treat that as an extension instead of adding it to core.

passion729:
@saeedvaziry 

Updated this PR and rewrote the branch history.

- Removed the bundled `rg` binary from core, including from the branch history.
- Rebased onto latest `main`; the PR is now scoped to the native FSEvents worktree watcher only.
- Fixed the app wiring so `VCSWorktreeAutoRefresher` is retained and started.
- Added coverage for linked-worktree `.git` resolution, including relative `gitdir:` paths.

Validation:
`mise exec swiftformat@0.60.1 swiftlint@0.57.1 -- scripts/checks.sh --fix` passes.

Could you please reopen this PR for review? If you still think this should be converted into an extension instead of living in core, please let me know and I’ll adjust the direction.

saeedvaziry:
@passion729 I cant re open the PR because you force-pushed. you need to open a new PR

passion729:
@saeedvaziry Thanks, that makes sense. I created a new PR from the cleaned-up branch: #712 

This version is rebased onto latest `main`, has a clean single-commit history, and removes the bundled `rg` binary entirely. The implementation is scoped to the native FSEvents worktree watcher only.

PR #606, opened 2026-05-29
feat: Low Memory Mode — flat RAM with many terminal tabs

dzulfikar08:
# Summary

Each Ghostty terminal surface eats ~80-100MB of RAM. Open 20 tabs and you're at ~2GB while most of them is idle but we don't want it closed. So I propose to use tmux that wraps it and aggressively kill when we switch Workspace, thanks to tmux, the process still running at the background, and switching back to the workspace will just reattach it instantly, result is steady 250-300 MB RAM.

## What this does

Adds an opt-in **Low Memory Mode** (Settings → Terminal → Performance) that wraps shells in persistent tmux sessions. When you switch workspaces, hidden terminal surfaces are destroyed after 3 seconds,  freeing that ~80MB per tab while tmux keeps the shells alive underneath. Switch back and it reattaches instantly, full shell state intact.

**Before:** 15 Workspaces → ~1.6 GB RAM
image

**After:** 15 Workspaces → ~230 MB RAM
image

## Trade-offs

These are documented in the settings UI so users know what they're signing up for:

- Smooth scrolling is replaced with tmux-style scroll (enters copy mode)
- Text selection is handled by tmux, not native Ghostty
- Existing terminals need to be reopened after enabling the setting
- Requires tmux (brew install tmux), setting is disabled with a warning if not installed
- Progress indicators and desktop notifications don't fire for hidden workspaces (surfaces are destroyed). Notifications still work for the active workspace.

Off by default, you have to explicitly turn it on.
image

## Testing

- [x] scripts/checks.sh passes (version mismatch pre-existing)

## Security Review
- [x] tmux binary detection with UI guard when not installed
- [x] Surface lifecycle: lazy creation, occlusion-based eviction, teardown cleanup
- [x] ShellEscaper for all tmux command arguments, write-once config with error logging


long2ice:
That's cool

saeedvaziry:
Thanks for the PR. I think Muxy shouldnt be using that much memory even in the normal mode

dzulfikar08:
> Thanks for the PR. I think Muxy shouldnt be using that much memory even in the normal mode

Thanks! I agree as it's muxy promise.

However this high usage is coming from the ghostty itself, opening multiple terminal of ghostty will give you same result (80-100MB). Muxy just managing many of them.

Do you prefer different approach?

saeedvaziry:
Your PR is solid and I like it. In fact it gives more terminal feel and I might even use it myself by default.

While testing, I've noticed the memory also goes high with this PR but is moderate.

And lastly, I don't think the numbers on the screenshot is the effective memory usage that Muxy has at the time. most of it seems like shared if you open the details.

saeedvaziry:
@dzulfikar08 would you be able to run a review and maybe a security review on the PR? I am getting critical warnings from opus and other agents.

dzulfikar08:
> @dzulfikar08 would you be able to run a review and maybe a security review on the PR? I am getting critical warnings from opus and other agents.

@saeedvaziry pushed the security review fix

saeedvaziry:
@dzulfikar08 it seems like no matter the low memory is ON or OFF, it will prevent the mobile app to use in-active tabs

dzulfikar08:
@saeedvaziry yeah just now i notice it too, i'm on it.

dzulfikar08:
image
fix pushed 
@saeedvaziry

saeedvaziry:
@dzulfikar08 after a lot of testing and reviewing, seems like this PR has architecture design issues. spawning lots of tmux sessions, writing some random characters into the terminal when using on mobile, and I think most importantly, making a dependency to tmux.

Would be nice if we can push more towards making the current implementation stable and more efficient

saeedvaziry:
Also some more critical points from reviews:

- Muxy/Services/TmuxCaptureService.swift:225: parseControlOutput assumes tmux %output payloads are base64. Actual
    tmux -C output is escaped text, e.g. %output %0 abc def\015\012. Hidden remote terminals will drop normal
    output and appear blank/stale.
- A new tmux process is spawned per keystroke / scroll / resize — TmuxCaptureService.swift:147,173,203

saeedvaziry:
@dzulfikar08 can you drop me a msg on discord? I have some ideas and would like to discuss

dzulfikar08:
@saeedvaziry yeah i agree, still lot of room to polish this. If only libghostty has API for dettach and attach, we will no need to use tmux. 

I just hit you on discord

saeedvaziry:
Thanks @dzulfikar08 for the PR. I introduced idle mode to free up memory which seems like working good. still a good idea to continue the idea from this PR but maybe in a different one.

Closing this now.

Thanks

PR #608, opened 2026-05-29
fix: race condition causing CLI socket commands to return no response after EOF

maplezzk:
## Problem
CLI commands like `muxy split-down` return "Error: no response from Muxy". The split succeeds, but the CLI never receives the response.

## Root Cause
Two issues in `NotificationSocketServer`:

1. **Premature dispose on EOF**: When `nc` sends a command and closes its write end, `readFromSession` detected EOF and called `disposeSession` before processing the buffered command line — the response was never even prepared.

2. **Async write race**: `processCommand` spawns a `Task` to invoke `SocketCommandHandler.handleRequest` on `@MainActor` and returns synchronously. After the command was processed, `readFromSession` would reach EOF (client already half-closed) and dispose the session — closing the fd before the async Task could enqueue and flush its response write.

## Fix
- Extracted `processBufferedLines` from `readFromSession` and call it before checking EOF, ensuring the command is always parsed before dispose
- Added `pendingClose` flag — on EOF, cancel only the read source while keeping the fd alive for pending writes
- Added `commandInFlight` flag — defer close when the async Task hasn't enqueued its write yet (buffer may be empty at EOF time)
- `flushWrites` closes the fd after draining when `pendingClose` is set; `closeSession` skips the fd when `pendingClose` is active

This fixes all CLI socket commands (`split-right`, `split-down`, `send`, `list-panes`, etc.) in one change.

## Files Changed
- `Muxy/Services/NotificationSocketServer.swift` — 25 insertions, 3 deletions

## Testing
- All 1265 existing tests pass
- Manually verified via local integration: built a signed release, ran `muxy split-down` / `muxy split-right` / `muxy list-panes` etc., all return correct responses
- Unit tests for this specific race condition are difficult to write without exposing internal socket/session management — the race involves socket half-close timing and async Task scheduling across the socket queue and MainActor

### DIFF
--- a/Muxy/Services/NotificationSocketServer.swift
+++ b/Muxy/Services/NotificationSocketServer.swift
@@ -20,7 +20,9 @@ final class NotificationSocketServer: @unchecked Sendable {
     final class ClientSession: @unchecked Sendable {
         static let droppedNotificationDisconnectThreshold = 100
 
-        let fd: Int32
+        var fd: Int32
+        var pendingClose = false
+        var commandInFlight = false
         var extensionID: String?
         var subscriptions: Set<String> = []
         var writeBuffer = Data()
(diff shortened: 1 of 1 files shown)


maplezzk:
Before fix:
image

maplezzk:
After fix:
image

PR #611, opened 2026-05-30
fix: cross-worktree pane lookup after split

maplezzk:
## Problem
`send` returns "pane not found" after splitting from a background worktree pane.

## Root Cause
1. `SplitReducer.splitArea` resolves workspace root via `activeKey`, which points to the active worktree — not the from-pane's worktree. The target areaID isn't found, so the split silently fails.
2. `waitForView` only passively polls `TerminalViewRegistry`. Panes in background worktrees may never be rendered by SwiftUI, causing a 3-second timeout.

## Fix
- `SplitReducer.splitArea`: search all workspace roots for the one containing the target areaID
- `waitForView`: eagerly create view + surface via `TerminalViewRegistry` + `materializeHeadless` when pane exists in state but has no view
- Added cross-worktree split test


maplezzk:
✅ Reviewed, tested, and linted — all checks pass.
- 1337 tests passing
- swiftformat + swiftlint clean
- build succeeds

maplezzk:
Local test results:
image
image

PR #636, opened 2026-06-01
fix: keep pane views mounted across worktree switches

maplezzk:
muxy send/send-keys/read-screen now materialize a pane's terminal surface on demand so they work against
  background worktrees/projects, and CLI replies are NUL-terminated so the client can tell an empty success from no
  response.


saeedvaziry:
@maplezzk my comment was AI of course because the PR summary itself was AI. I'd appreceate we keep this for humans for further communications.

I am already checking the chagnes...

maplezzk:
@saeedvaziry Me too.I'm not sure if what the AI said is correct, so I can only ask it to fix it and then test it myself

saeedvaziry:
@maplezzk would you solve the conflict so I can test?

maplezzk:
@saeedvaziry Processing

maplezzk:
@saeedvaziry After the adjustment, other issues arose, and I still need to make further adjustments

maplezzk:
@saeedvaziry It should be fine now, try again

saeedvaziry:
I will solve the conflicts and test

maplezzk:
@saeedvaziry Just fixed

maplezzk:
@saeedvaziry Please recheck

saeedvaziry:
@maplezzk this PR is still agaist the optimization I made. it tries to set visibility false to the background tabs to just make them work. I see this as a hacky solution.

However I do accept the problem is real. Looking into it and will update this PR. please dont push anything.

saeedvaziry:
I tested this locally and works. let me know if works on your side as well

maplezzk:
@saeedvaziry OK, let me test it

maplezzk:
@saeedvaziry found two issues
1. The read-screen will report an error 'no response from Muxy'. Steps are as follows:
$ /tmp/repro-muxy-pane-not-found.sh
→ Waiting 3 seconds — please switch the current worktree to background in Muxy...
✓ 3 seconds elapsed. Running muxy CLI inside the backgrounded worktree now
→ Current pane: CCFD8439-100F-481B-A6E7-446BEBB3DC53

→ Step 1: muxy split-right --from $MUXY_PANE_ID
✓ split succeeded, returned pane ID: 3DB98F04-79A3-4327-87CA-12EA0D0871A3

→ Step 2: muxy send --pane $PANE 'ls -l'
✓ send succeeded (unexpected — bug is fixed)

→ Step 3: muxy send-keys --pane $PANE Enter
✓ send-keys succeeded (unexpected — bug is fixed)

→ Step 4: muxy read-screen --pane $PANE --lines 20
✗ read-screen failed: Error: no response from Muxy (exit=1)

2. When I switch back to the background pane, it seems that the result of the executed command ls -l is not rendered, instead showing a blank pane

Foreground:
image

Background:
image

saeedvaziry:
@maplezzk those issues fixed already. you need to hit Install Cli again and then test it

maplezzk:
@saeedvaziry Issue 1 is resolved, but Issue 2 still persists.
image

maplezzk:
This is the script I executed:
[repro-muxy-pane-not-found.sh](https://github.com/user-attachments/files/28680926/repro-muxy-pane-not-found.sh)

saeedvaziry:
@maplezzk I think thats expected tho. if a terminal tab goes to sleep. its content will be gone. 

However I see the point we should save a snapshot and return that instead so I'll handle that in another PR instead.

meanwhile you can totally disable tabs going to sleep in Settings -> Terminal

PR #653, opened 2026-06-03
fix(remote): Fix blank terminal when a terminal tab is created from the mobile app

hlouis:
## Summary

  Creating a terminal tab from the mobile app showed a blank screen — the desktop tab existed but the shell produced no output until I focused the desktop window, at which point the shell initialized and the prompt appeared.

  When a pane is created headlessly for a remote client, the surface was created at a 1×1 size, so libghostty never computed cell metrics. `applyPTYSize` needs those metrics to size the PTY and bailed out when they were missing, so the PTY was never sized to the client's requested grid and the shell produced no output until the desktop rendered the surface for real.

  ## Testing

  - [x] Tested manually on macOS and iOS (use store version)


saeedvaziry:
@hlouis I made some changes already on the main branch and cant reproduce this in the last beta. let me know if you still can.

Closing the PR until then

PR #660, opened 2026-06-04
Add configurable tab header sizing

Thinkscape:
## Summary
- Add a `Tab header size` setting in Interface/Appearance with `Small`, `Medium`, and `Full-width` presets.
- Store preset widths as 200px and 400px, while `Full-width` clears the stored max width so tabs stretch.
- Keep direct JSON/config edits flexible by accepting any finite nonnegative pixel width; `null` and `0` both mean full-width.
- Let the tab strip expand across the titlebar when uncapped and apply the configured max width when present.

## Verification
- `swift test --filter SettingsJSONStoreTests`
- `swift build --product Muxy --skip-update`
- `git diff --check`
- Launched a locally built debug app bundle, selected a seeded sidebar project, verified the single tab stretches in `Full-width`, verified the settings UI shows `Tab header size` with `Small`, `Medium`, and `Full-width`, and verified selecting `Small` caps the visible tab handle.

## Before
Screenshot 2026-06-04 at 3 12 22 pm

## After

Screenshot 2026-06-04 at 3 16 41 pm

Screenshot 2026-06-04 at 3 16 24 pm


Thinkscape:
That was a major annoyance for me as I couldn't see the harness' session names...

saeedvaziry:
@Thinkscape Thanks for the PR but I disagree with this change. it looks weird to me honestly that size of the tab

Thinkscape:
> @Thinkscape Thanks for the PR but I disagree with this change. it looks weird to me honestly that size of the tab

That's why I made it configurable...  see the third screenshot.

Thinkscape:
> @Thinkscape Thanks for the PR but I disagree with this change. it looks weird to me honestly that size of the tab

Again. it feels like your particular biases get in the way of this open source project.

- MacOS Terminal
- Iterm2
- Ghostty 
- Warp

All of those are orders of magnitude more popular terminal emulators than muxy, and **all of them have by default flex tab handle widths**. I've made it configurable so that you can keep your one-person preference intact, while allowing others to have behaviour aligned with literally every other terminal app.

saeedvaziry:
Its not biased. I would call it standard.

I will check how other terminals implemented this.

Thinkscape:
> Its not biased. I would call it standard.

100% biased. Like proven above, everything else in this category does tabs differently, and I've provided a configurable way to suit your unique taste, while making it usable to everyone else, and you decided to close the PR ... can't be any more biased than that.

Thinkscape:
Any news on that?

PR #706, opened 2026-06-07
fix: split cli from calling pane via MUXY_PANE_ID

maplezzk:
`muxy split-right` / `split-down` now default to splitting the **calling pane** (via the already-injected `$MUXY_PANE_ID`) instead of the focused pane, so splits land where you typed the command, not where you last clicked. `--from` still overrides. Backward compatible: outside a Muxy pane `$MUXY_PANE_ID` is unset, and the old focused-pane fallback kicks in (though Muxy CLI is not meant to run outside a Muxy pane).


maplezzk:
@saeedvaziry A bit strange, I didn't modify the Swift code, so why did it fail? And it runs successfully on my local machine

maplezzk:
@saeedvaziry Recheck please. Why isn't auto-recheck enabled? Is it due to GitHub rate limiting or some other reason?

PR #716, opened 2026-06-08
SSH remote project management with Keychain auth

### LINKED ISSUE #438: [Feature Request] Remote SSH connections - manage terminal sessions on remote Linux servers
### Problem

Currently muxy only works with local terminal sessions. Many developers work with remote Linux servers and need to manage terminal sessions on those machines. Right now the only option is to SSH manually in each tab, which means losing muxy’s project organization, tab management, and multiplexing features for remote work.

### Proposed Solution

Add first-class remote connection support via SSH, similar to how OpenAI Codex handles remote connections. Specifically:

1. SSH host configuration: Let users define remote hosts in muxy’s project config (or auto-discover from ‎`~/.ssh/config`). Each host entry includes hostname, user, identity file, and optionally a port.

2. Remote project folders: When adding a project, allow selecting a remote host + directory path. Tabs opened in that project would automatically SSH into the remote machine and cd to the project directory.

3. Session persistence: Keep remote sessions alive across disconnects. If the network drops temporarily, muxy should reconnect and resume the session rather than losing the terminal state.

4. Per-project host binding: A project can be bound to a specific remote host, so all new tabs/panes in that project automatically connect to the remote machine without the user needing to type ‎`ssh` each time.

5. Connection status indicator: Show connection state (connected / reconnecting / disconnected) in the UI for remote projects.

### Additional Context

Reference: OpenAI Codex’s remote connections feature (https://developers.openai.com/codex/remote-connections)

This would make muxy a strong option for developers who split work between local and remote environments.

EkkoG:
## Summary

Add SSH remote host management so users can save server connections and open remote directories as first-class projects in the sidebar, with automatic password authentication via macOS Keychain.

## Changes

- RemoteHost model with SSH connection config storage
- RemoteProjectConfig to associate remote paths with hosts
- Project.Kind enum to distinguish local and remote projects
- KeychainSSHHelper for macOS Keychain password storage
- RemoteHostStore with JSON persistence
- SSHConfigParser for ~/.ssh/config import
- AddRemoteHostSheet / AddRemoteProjectSheet / SSHConfigImportSheet
- SSHHostManagementPanel for host CRUD
- TabArea SSH command injection with ControlMaster multiplexing
- TerminalOmnibox integration for quick host connection
- Connection failure detection with inline reconnect overlay
- Session restore skips remote panes

## Testing

- [x] swift build passes
- [x] swift test passes (1297 tests)

Closes #438

SCR-20260608-odzy
SCR-20260608-oeor-2


EkkoG:
Implemented with DeepSeek V4 Pro vibe coding.

andrejpalovic:
This functionality would be great. I’m not sure whether this is the right implementation, but in any case, when I have both a server and a workstation, the current workflow feels quite limited: if I want to start an agent on the server and work with it, I can only really do that through the mobile app.

If there were some way to connect multiple muxy instances, that would be great.

EkkoG:
Update: this change has been reimplemented using Swift NIO. The SSH path now uses NIO handlers and event-loop driven IO, replacing the prior blocking/script-based flow. It also fixes key host matching and connection lifecycle handling: non-blocking EOF/error handling, retry on temporary write errors, and deterministic connection cleanup when sessions close or fail.

saeedvaziry:
@EkkoG I am still not a fan of this feature and I dont think Muxy should have it by default.

However I am ready to expose enough APIs via extensions so you can turn this feature into an extension API instead.

Would you be able to?

EkkoG:
Got it. I’m happy to adapt once the extension APIs are available. Until then I’ll keep maintaining a fork version to avoid blocking the current release path.

saeedvaziry:
Cool. would be nice if you tell me what APIs can support this. I will also check from the PR itself

saeedvaziry:
This PR will enable you to even ship your own sidebar which has remote SSHs as well.

https://github.com/muxy-app/muxy/pull/724

EkkoG:
I’ve reconsidered, and I’m leaning toward making SSH a built-in core capability in the main app rather than a plugin feature. This isn’t a preference issue, it’s a product boundary issue:

  1. SSH touches the core session orchestration path—session lifecycle, terminal I/O, reconnect/EOF handling, project restore, and integration points like sidebar/omnibox entries. If this is moved to a plugin, the key
     behavior becomes fragmented across external extensions.

  2. A usable implementation requires exposing more than a few UI hooks: host configuration, connection management, Keychain-backed authentication, session events, error/retry handling, remote project modeling, and
     command injection for startup.

  3. That means SSH should first be a built-in, foundational capability, while plugins should build on top of it for custom UI or workflow extensions (for example, custom sidebars), not replace the core.
  4. Making this feature plugin-based is significantly more complex: it introduces extension lifecycle constraints, IPC/serialization overhead, permission and trust boundaries, and cross-extension compatibility issues,
     which makes debugging and reliability much harder.

  5. This also improves maintainability: consistent behavior and security boundaries, better stability, and a cleaner API surface for the extension ecosystem instead of each extension reimplementing transport logic.

saeedvaziry:
Lets discuss this for version 1.1.0

I dont want to add SSH blindly. When we talk remote, it has to be truely remote. Even a mac-to-mac via Muxy API.

EkkoG:
What is truely remote?

saeedvaziry:
SSH + Mac-to-Mac remote connection (no ssh) + git api, files api, all extensions api, ...

PR #727, opened 2026-06-09
resume claude code sessions when restoring workspaces

Hude06:
Restored tabs always relaunched as bare shells, so Claude Code sessions running at quit were lost on restart.
Snapshots now record a resume command when a known agent is the pane's foreground process, and restore replays it (`claude --continue`) in the saved working directory.
No UI change; covered by new snapshot/restore tests, and degrades to a plain shell when there is nothing to resume.


saeedvaziry:
Closing in favor of #725 

I dropped the feature and provided api for extensions instead. you can make an extension for this

Hude06:
Okay interesting, I will create an extension for this. Thank you!

PR #736, opened 2026-06-10
Fix icon corner radius to match macOS squircle style

Hude06:
Use `style: .continuous` on project icon `RoundedRectangle` shapes so corners match the macOS app icon squircle curve. Also corrects the active selection border radius from `scaled(11)` to `radiusMD + scaled(3)` so it sits proportionally parallel to the icon it wraps.


Hude06:
Heres the pic
CleanShot 2026-06-10 at 13 01 48@2x
Just wanted them to be the same, if you wanted it a specific way that is fine

PR #739, opened 2026-06-10
Add remote (SSH) workspaces

### LINKED ISSUE #438: [Feature Request] Remote SSH connections - manage terminal sessions on remote Linux servers

saeedvaziry:
Closes #438


EkkoG:
I pulled this PR locally and tried it out. I agree that remote operations need a consistent execution target, but I don’t think that target necessarily has to be the workspace itself. In practice, a mixed setup is
  pretty common: for example, a local frontend project, a remote backend project, and a remote agent session can all belong to the same working set. If SSH is modeled strictly as a workspace property, users have to
  switch workspaces just to switch execution targets, which breaks that workflow apart. I think it would be better to decouple workspace organization from remote execution context: let the workspace remain a container
  for related projects, and let each project or active session carry the execution target that determines where files, git, commands, and extension actions resolve. That still preserves consistency, but avoids making
  mixed local/remote work unnecessarily awkward.

saeedvaziry:
@EkkoG I see your point. I like the workspace one but its easier and more flexible in project level remote server probably.

saeedvaziry:
I think the best option would be both solutions. as a workspace and as a project.

PR #753, opened 2026-06-12
Experimental native SSH support via swift-nio-ssh

EkkoG:
## Summary

Experimental change replacing ssh-cli with swift-nio-ssh for password login support and more controllable SSH connection management.

## Changes

- replace ssh-cli with swift-nio-ssh
- support password-based SSH login
- move SSH session lifecycle into app-managed code

## Testing

- [x] scripts/checks.sh passes


saeedvaziry:
@EkkoG can you explain a bit the vision using `swift-nio-ssh`?

EkkoG:
The main goal is to move SSH from an external `ssh` process into Muxy itself.

  Compared with the CLI-based approach, this gives us direct control over authentication, session lifecycle, and pane
  integration, and it also makes password-based login a first-class path.

  It also gives us a better foundation for follow-up flows that are awkward with an external `ssh` process, such as reusing the
  authenticated session context for remote bootstrap/setup steps and building a more consistent password experience around
  things like `sudo`.

  That said, this PR is intentionally experimental. It does not assume that `swift-nio-ssh` is automatically better than the
  CLI path in every respect. The tradeoff is that we now own more of the SSH surface ourselves, especially around key handling,
  shell compatibility, and recovery.

  So the short-term goal is to validate whether this approach is a better fit for Muxy’s remote-project workflow.

saeedvaziry:
@EkkoG For this experimental feature we can ship ssh cli and swift-nio-ssh at the same time and default to ssh cli in the settings

EkkoG:
@saeedvaziry Sounds good to me. This approach is definitely better.

saeedvaziry:
@EkkoG will you apply this?

> @EkkoG For this experimental feature we can ship ssh cli and swift-nio-ssh at the same time and default to ssh cli in the settings

or you have already applied it?

EkkoG:
Thanks for checking!
I haven’t applied the dual‑SSH setup yet.
I’ll update the PR to ship both ssh‑cli and swift‑nio‑ssh, with ssh‑cli as the default, as discussed.

saeedvaziry:
@EkkoG closing this as stale. feel free to open a new one

PR #762, opened 2026-06-15
Add per-worktree AI agent status to the extension API

loeck:
## Summary

Adds a per-worktree AI agent status to the extension API. Extensions can react to an agent's lifecycle (`working` / `waiting` / `idle`) via the new `agent.status` event and hydrate current state on load with `muxy.agents.list()`, gated behind a new `agents:read` permission. Status is driven by each provider's hooks Claude Code, Cursor, Codex, Droid, OpenCode, Pi and aggregated per worktree (most active wins, ties break on recency).

## Changes

- Add `AgentStatusStore` aggregating pane statuses per worktree, broadcasting only on change (`AgentStatusStore.swift`).
- Emit `agent_status` from every provider's hook/plugin (`muxy-*-hook.sh`, `opencode-muxy-plugin.js`, `muxy-pi-extension.ts`); `ClaudeCodeProvider`/`DroidProvider` refactored to a shared `hookEvents` table.
- Parse and route `agent_status` socket messages, accepted only from the host hook, not extensions (`NotificationSocketServer.swift`).
- Add the `agents.list` verb + `agents:read` permission and dispatch wiring (`MuxyAPI.swift`, `MuxyAPIDispatcher.swift`, `Extension.swift`, `ExtensionEvent.swift`); expose `muxy.agents.list()` in both JS bridges.
- Emit `idle` and drop state when the last agent pane closes (`ExtensionEventEmitter.swift`), and also when the agent's foreground command exits while the pane stays open (`GhosttyRuntimeEventAdapter.swift`) — so an agent quit via Ctrl-C/`/exit`/crash (no `Stop` hook) never leaves a stale status.
- Enrich the `notification.posted` payload (`worktreeID`, `worktreePath`, `source`, `body`) + add `MuxyNotification.Source.key`, reused by the status dispatch.
- Update docs, manifest schema and the extension skill; add unit tests.

## Provider coverage

Reported states depend on what each CLI's hooks expose:

| Provider | `working` | `waiting` | `idle` |
| --- | --- | --- | --- |
| Claude Code, Droid, OpenCode | ✓ | ✓ | ✓ |
| Pi | ✓ | — | ✓ |
| Cursor | — | ✓ | ✓ |
| Codex | — | — | ✓ |

## Testing


saeedvaziry:
@loeck is this currently only supports Claude Code? if yes, we need to add support to all our supported providers

loeck:
> @loeck is this currently only supports Claude Code? if yes, we need to add support to all our supported providers

Good catch indeed ! I will check how I can do that for all providers 👌

saeedvaziry:
I'd also apreceate a video recording on an example extension to see the real use case

PR #764, opened 2026-06-16
Add configurable project sorting and running indicator

long2ice:
Add sort modes (manual, name A–Z/Z–A, recently active, date added), switchable from the sidebar menu and Projects settings.
Dragging in a non-manual mode persists the order and switches back to manual.
Project icons now show a spinner while a terminal reports progress, then the existing completion dot.


long2ice:
image
image
image

PR #567, opened 2026-05-26
Fix Pi notification pill showing wrong provider name

maxschulmeister:
Fixes muxy-app/muxy#484

## Summary
- Resolve notification toast and panel titles from `AIProviderRegistry` for known socket types (e.g. `pi` → "Pi") instead of the raw hook title field.
- Map socket payloads to provider display names at ingest time in `NotificationSocketServer`.
- Add `AIProviderRegistryTests` and document Pi in notification docs.

## Test plan
- [x] `swift test --filter AIProviderRegistryTests`
- [x] Manual socket test: toast pill and notification panel show **Pi** for `pi|…|Claude Code|…`
- [ ] CI on upstream

Related: https://github.com/muxy-app/muxy/issues/484

Made with [Cursor](https://cursor.com)

PR #796, opened 2026-06-22
Use app-local Pi wrapper and correct Pi notification labels

maxschulmeister:
Updates Pi notifications to use an app-local wrapper instead of copying Muxy's extension into Pi config.\nCorrects the Pi provider label shown in notification pills and preserves current notification behavior.\nValidation: scripts/checks.sh --fix; runtime socket notification smoke test.


saeedvaziry:
We have had multiple Pi notification PRs with different integration models:

- #529/#564 installed and registered the Pi extension through Pi’s config.
- #566 proposed an app-local wrapper instead, but was closed.
- This PR revives the wrapper model and removes the existing Pi config integration path.

Before merging, we should decide which Pi integration architecture we want long-term: mutate/register in Pi config, or app-local wrapper/PATH shadowing. Otherwise we keep replacing one approach with another and accumulating edge cases.

maxschulmeister:
#566 and #567 were both filed by me and this PR combines both as requested.
i think #529 and #564 are fine, as they have been merged already, but they introduce 2 caveats.

1) Notifications can show up with a wrong title because they use the raw hook title field instead of AIProviderRegistry
2) A less invasive approach in regards to the pi extension would be preferable and more stable. So instead of copying the pi extension I suggest to use a wrapper that directly loads the extension.

saeedvaziry:
@maxschulmeister why changing the checks script and swift version? I think thats breaking the checks in this PR

saeedvaziry:
Closing as stale. feel free to open a new one

PR #835, opened 2026-06-28
Make terminal cmd-click file open behavior configurable

### LINKED ISSUE #732: In version 1.0 : Open a file will no longer create a new tab
### Problem

In 1.0 version extension,it show different file in one same tab
In previous version feature, click file will show it in a new tab

### Proposed Solution

maybe this feature can be a choice that user can edit

### Additional Context

_No response_

mvanhorn:
## Summary

Closes #732. Cmd-clicking a file path in the terminal always falls back to launching an external editor, with no way to keep files inside Muxy. This adds a Terminal setting to pick between that behavior and using a registered in-app opener only.

## Changes

- New "Files" picker in Terminal settings: "Open in external editor" (the current default) or "Open with in-app opener only".
- `TerminalPane` now reads the setting on cmd-click and OSC8 file links. When in-app-only is chosen and no opener is registered, it shows a notice and skips the external editor instead of launching it.
- Registered the key for settings search and JSON validation, and added unit tests for the behavior model.

## Testing

- [x] `swiftformat --lint` and `swiftlint --strict` clean on the changed files
- [x] Added `FileOpenBehaviorTests` covering the setting model and its default

I have not run the full `scripts/checks.sh` build or a manual macOS pass yet, so please confirm those before merging. The default keeps today's behavior, so existing users see no change.

## Screenshots

New Terminal settings row "Cmd-click a file path" with the two options.

### DIFF
--- a/Muxy/Models/GeneralSettingsKeys.swift
+++ b/Muxy/Models/GeneralSettingsKeys.swift
@@ -6,8 +6,8 @@ enum GeneralSettingsKeys {
 }
 
 enum TerminalFileOpenBehavior: String, CaseIterable, Identifiable {
-    case externalEditor = "External editor"
-    case inAppOpener = "In-app opener"
+    case externalEditor = "Open in external editor"
+    case inAppOpener = "Open with in-app opener only"
 
     static let defaultBehavior: TerminalFileOpenBehavior = .externalEditor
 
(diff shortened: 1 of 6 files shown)


saeedvaziry:
@mvanhorn I think this is solved here https://github.com/muxy-app/muxy/pull/811

PR #838, opened 2026-06-28
Fix CLI socket reliability and headless browser screenshots

saeedvaziry:
Drains the accept backlog so the CLI listener no longer wedges after a couple of uses, and delivers large replies reliably by keeping the client session alive through the write retry.

Browser `screenshot` now captures via `createPDF`, so it works headlessly on background tabs; DEBUG builds bind their own socket so `swift run Muxy` no longer collides with the installed app.

Skill/docs/CLI help updated for headless browsing and screenshots.


saeedvaziry:
Verified end-to-end: a headless screenshot of a **background** browser tab, driven from a terminal tab via `muxy browser screenshot` (CLI never switched to the browser tab). Stress-tested 120+ captures with 0 failures.

Captured PNG: `Example Domain` page rendered off-screen, 3150×1910.

PR #865, opened 2026-07-05
Add cancellable async extension exec API

wjchoi87:
## Summary

Adds a cancellable async command execution API for extensions so long-running native commands can be started without blocking the JS queue and cancelled when superseded. This keeps the existing synchronous `muxy.exec(...)` API unchanged for backward compatibility.

## Changes

- Added `muxy.execAsync(...)`, returning `{ id, result, cancel() }`
- Added native cancellable exec job tracking with `jobID -> Process`
- Terminate running processes on cancel, with timeout/cancel race protection
- Preserve existing `muxy.exec(...)` behavior, permissions, timeout handling, and result shape
- Return identifiable cancellation errors via `error.code === "cancelled"` and `error.cancelled === true`
- Added tests for normal async completion, cancellation, cancel-after-completion, timeout/cancel race, and JS queue delivery
- Updated extension script docs and extension skill guidance

## Testing


saeedvaziry:
@wjchoi87 everytime I review this with opus or gpt 5.6 sol I get lots of critical and warnings. would you have a look?

wjchoi87:
@saeedvaziry I have reviewed the critical issues and warnings and fixed any potential issues.
If there are still any problematic elements, please let me know which ones.

PR #871, opened 2026-07-07
Fix terminal scroll after window regains focus

### LINKED ISSUE #845: Scroll failed after lost or re-gain focus
### Description

when Muxy lost focus(even re-gain focus), mouse scroll in OpenCode not work. (for modern OS, mouse cursor scroll on unfocused app will act normal.
when re-gain focus, it still not work until you click terminal area one more time

### Steps to Reproduce

1. Open Muxy
2. switch focus to other app or screen
3. mouse over Muxy and scroll, not working
4. switch back to Muxy, scroll still not working
5. click one more time to Muxy terminal area, scroll working

### Muxy Version

1.1

### macOS Version

26.5.1

Hude06:
Fixes #845

### DIFF
--- a/Muxy/Views/Terminal/GhosttyTerminalNSView.swift
+++ b/Muxy/Views/Terminal/GhosttyTerminalNSView.swift
@@ -279,6 +279,10 @@ final class GhosttyTerminalNSView: NSView {
             NotificationCenter.default.removeObserver(observer)
             keyWindowObserver = nil
         }
+        if let observer = keyWindowResignObserver {
+            NotificationCenter.default.removeObserver(observer)
+            keyWindowResignObserver = nil
+        }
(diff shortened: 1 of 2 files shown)


saeedvaziry:
I think this is close, but I’m not sure it fully fixes the user case yet.

  From what I can tell, when Muxy loses focus and comes back, we may still think the terminal is already
  focused, so this new “window became active” hook might not actually tell the terminal to refresh its
  focus state. That means the scroll issue could still happen after switching away and back.

  Can you also handle the opposite event, when the window loses focus, so the terminal state is reset
  before it becomes active again? A small test around that flow would be helpful too.

Hude06:
Yes I see this thank you for the comment, I am working on a fix rn. Thanks!

PR #868, opened 2026-07-06
fix split divider hit area

wjchoi87:
## Summary

Reduce the resize handle hit area so terminal text near dividers is easier to click and select, especially the first character next to a divider.

Fixes muxy-app/muxy#737

## Changes

- Reduced resizeHandleHitArea from 18pt to 10pt
- Updated the metric coverage test to match the new value

## Testing

- [x] `scripts/checks.sh` passes
- [x] Tested manually on macOS

PR #887, opened 2026-07-12
fix resize handle layout spacing

wjchoi87:
## Summary

Follow-up to #868 that fixes the remaining layout gap caused by resize handle hit areas consuming panel space.

## Changes

- Keep split dividers at 1pt while moving the larger drag target into a non-layout overlay
- Restore split container sizing so panes no longer lose extra space around each divider
- Preserve the wider grab area for right-side panel handles without adding visible spacing

## Testing

## Screenshots

image


saeedvaziry:
@wjchoi87 this kinda went back to the problematic part.

Try to open the filetree or git panel and then in the terminal generate content to make terminal scroll appear. then try to grab the terminal scroll.

PR #893, opened 2026-07-13
feat: add configurable global hotkey workspace

vipinchan:
## Motivation

Muxy currently requires switching to its main window before using a terminal. This adds an optional hotkey workspace that can appear over the current macOS Space, including another app's fullscreen Space, without moving or rehosting the normal Muxy window.

## What changed

- Adds a dedicated, nonactivating hotkey workspace window.
- Supports Double Command, Double Control, and Double Option triggers.
- Adds a configurable 100–1000 ms double-tap interval (300 ms by default).
- Optionally uses the trigger to both show and hide the workspace.
- Preserves normal and overlay-fullscreen presentation across hide/show.
- Keeps the normal Muxy window and hotkey workspace tabs/workspace state independent.
- Leaves the global hotkey disabled by default; enabling it installs the global event monitor.
- Creates the hotkey workspace lazily on the first successful trigger.

## Settings

Adds **Settings → Shortcuts → Global Hotkey**:

- Enable Global Hotkey
- Trigger
- Double Tap Interval
- Toggle to Hide

The preferences use Muxy's existing `@AppStorage` / `UserDefaults` pattern, are registered in `SettingsCatalog`, and participate in `settings.json` / `default-settings.json` sync and validation. **Reset All** on the Shortcuts page also restores the Global Hotkey defaults.

Modifier double-tap detection remains separate from `KeyBindingStore` because `KeyCombo` does not model modifier-only down/up timing or double-tap intervals. Using the selected modifier with a normal key, or holding it too long, does not count as a tap.

The Trigger control uses a theme-aware `Menu` so its label explicitly follows Muxy's custom Settings foreground/background colors in dark themes.

## Implementation notes

The workspace uses a dedicated `NSPanel` and `AppState`. It reuses Muxy's project persistence and shared services, but does not mutate the normal main window.

Overlay fullscreen is used so the workspace can remain on the current Space instead of creating or switching to a separate native fullscreen Space.

## Testing

Covers:

- modifier double-tap recognition and interval handling
- normal key-combination and long-press rejection
- Command, Control, and Option trigger persistence
- opt-in defaults, interval clamping, and preference reset
- Settings JSON persistence and invalid trigger/interval rejection
- hotkey window shortcut scope recognition

Repository Checks run SwiftFormat, SwiftLint, `swift build --build-tests --quiet`, and the full `swift test --quiet` suite.

## Manual validation

Validated on macOS for:

- presentation over normal and fullscreen apps
- show/hide from another application's Space
- overlay-fullscreen state preservation
- tabs, panes, sidebar, and workspace switching
- live Global Hotkey settings changes

### DIFF
--- a/Muxy/Services/Keyboard/GlobalHotkeyPreferences.swift
+++ b/Muxy/Services/Keyboard/GlobalHotkeyPreferences.swift
@@ -1,0 +1,74 @@
+import Foundation
+
+enum GlobalHotkeyTrigger: String, CaseIterable, Identifiable {
+    case doubleCommand
+    case doubleControl
+    case doubleOption
+
+    var id: String { rawValue }
+
+    var title: String {
+        switch self {
+        case .doubleCommand:
+            "Double Command"
+        case .doubleControl:
+            "Double Control"
+        case .doubleOption:
+            "Double Option"
+        }
+    }
+}
+
+enum GlobalHotkeyPreferences {
+    static let enabledKey = "muxy.globalHotkey.enabled"
+    static let triggerKey = "muxy.globalHotkey.trigger"
+    static let doubleTapIntervalMillisecondsKey = "muxy.globalHotkey.doubleTapIntervalMilliseconds"
+    static let toggleToHideKey = "muxy.globalHotkey.toggleToHide"
+
+    static let defaultEnabled = false
+    static let defaultTrigger = GlobalHotkeyTrigger.doubleCommand
+    static let defaultDoubleTapIntervalMilliseconds = 300.0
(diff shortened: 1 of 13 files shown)


saeedvaziry:
@vipinchan I didnt review this because its in Draft mode. I see no activity. is it ready?

saeedvaziry:
feel free to re-open

vipinchan:
I rebuilt this on the current `main` branch and moved the updated implementation to #1092. The new PR keeps Global Workspace separate from Quick Terminal while using Muxy’s current shortcut architecture.

PR #912, opened 2026-07-17
fix: prevent Esc from exiting full screen when Terminal Omnibox is open

agustinzamar:
Adds a local key-event monitor to TerminalOmniboxOverlay that intercepts Esc before AppKit's full-screen shortcut consumes it.

- Reusable OverlayEscapeKeyMonitor modifier + unit tests
- Manual check: full-screen + Esc now closes the omnibox and keeps full-screen mode


saeedvaziry:
@agustinzamar app is crashing when opening omnibox

agustinzamar:
Should be working now

PR #942, opened 2026-07-24
Fix stale default branch resolution

pedrommone:
## Summary

- Resolve the default branch from the remote HEAD before consulting local metadata.
- Preserve local and GitHub CLI fallbacks when the remote is unavailable.
- Cover remote HEAD parsing with regression tests.

## Verification

- `scripts/checks.sh --fix`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Git repository detection by identifying the default branch directly from the remote repository when available.
  * Preserved fallback behavior when remote branch information cannot be determined.

* **Tests**
  * Added coverage for successful default-branch detection and responses without symbolic branch information.

### DIFF
--- a/Muxy/Services/Git/GitRepositoryService.swift
+++ b/Muxy/Services/Git/GitRepositoryService.swift
@@ -880,6 +880,16 @@ struct GitRepositoryService {
     }
 
     private func resolveDefaultBranch(repoPath: String) async -> String? {
+        let remoteHead = try? await runGit(
+            repoPath: repoPath,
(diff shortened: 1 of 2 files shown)


pedrommone:
@saeedvaziry can you check this PR? I'm being with this annoying bug.

pedrommone:
Addressed the review feedback in 9b3da96. Added real-Git regression coverage for a stale local origin/HEAD being overridden by the remote symbolic HEAD, plus the non-symbolic remote HEAD fallback. All project checks pass.

PR #953, opened 2026-07-26
Keep terminal processes running when the window is backgrounded

mutallipp:
set_occlusion(false)、pty backpressure / buffer fills、child process blocks on write、freezes.
adds a toggle、defaults to on / enabled、keeps background panes reported as visible to Ghostty.
https://github.com/ghostty-org/ghostty/issues/5492.

## Summary by CodeRabbit

* **New Features**
  * Added a Background section in Terminal settings with a “Keep processes running in the background” toggle.
  * The setting is persisted and includes guidance about its impact on hidden rendering and I/O.

* **Bug Fixes**
  * Terminal visibility/occlusion now updates to match the Background setting, including when the toggle changes.

* **Tests**
  * Added tests covering defaults, persistence, notification behavior, and effective visibility logic.

### DIFF
--- a/Muxy/Extensions/Notification+Names.swift
+++ b/Muxy/Extensions/Notification+Names.swift
@@ -20,6 +20,7 @@ extension Notification.Name {
     static let focusBrowserSettings = Notification.Name("MuxyFocusBrowserSettings")
     static let focusQuickTerminalShortcut = Notification.Name("MuxyFocusQuickTerminalShortcut")
     static let quickTerminalEnabledDidChange = Notification.Name("MuxyQuickTerminalEnabledDidChange")
+    static let backgroundActivityKeepActiveDidChange = Notification.Name("MuxyBackgroundActivityKeepActiveDidChange")
     static let windowFullScreenDidChange = Notification.Name("MuxyWindowFullScreenDidChange")
     static let toggleSidebar = Notification.Name("MuxyToggleSidebar")
     static let toggleAppLayout = Notification.Name("MuxyToggleAppLayout")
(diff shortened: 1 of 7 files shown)


saeedvaziry:
@mutallipp this should already work. how can I reproduce the issue?

mutallipp:
@saeedvaziry     Here's my real-world repro:

   1. In Muxy, start an AI coding agent that streams output continuously
      (e.g. Claude Code, or any LLM CLI that streams tokens + tool calls)
   2. While it's working, switch to another tab in Muxy (or Cmd+Tab to
      another app)
   3. The agent appears to "fall asleep" — it stops producing output entirely
   4. Switch back to the terminal tab → it immediately resumes, dumping all
      the output it buffered

   It looks like the process is paused/sleeping, but it's actually blocked
   on write() because the pty buffer is full.

   The cause is in `applyOcclusionState()`:

       ghostty_surface_set_occlusion(surface, isPaneVisible && isWindowVisible)

   When the pane is hidden (tab switch) or the window loses focus (Cmd+Tab),
   Ghostty stops draining the pty → backpressure → the agent process blocks
   on its next write and effectively freezes.

   For long-running AI agents this is a big problem — a task that should
   take 2 minutes silently stalls for 10+ minutes until you happen to look
   at the terminal again. You don't even know it stopped.

   This doesn't happen in standalone Ghostty (it doesn't occlude on tab
   switch the same way), only when embedded in Muxy via the surface API.

saeedvaziry:
But does it stops working? have you tried to disable the auto sleep feature?

image

to turn this toggle off.

also it is only important to keep them working in the background without hiding them visually only to save memory and cpu to remain lightweight

saeedvaziry:
I believe this should be fixed in #961

Will drop a beta release soon. feel free to reply to reopen this if it hasnt fixed the issue

PR #961, opened 2026-07-27
fix: don't sleep while process is running

### LINKED ISSUE #894: Running long working session with omp, the tab would eventually go to sleep
### Description

I'm trying out https://omp.sh/ for my iOS coding. Sometimes omp will run a xcode test that takes long time to compile and run, if I don't have that tab active in Muxy, that tab will go to sleep after a while, results the test interrupted. This doesn't really happen when I was using Codex or Claude Code. 

### Steps to Reproduce

1. Open Muxy
2. Open omp
3. Run a task without any immediate feedback, like running a long test
4. The tab goes to sleep when it's not selected

### Muxy Version

1.2.1

### macOS Version

26.5

dergachoff:
Fixes #894.

I use Muxy for coding agents and run a few of them in bg tabs. Noticed that Muxy kept killing omp (oh my pi) 5 mins after I switch away. Tab goes offline, pty is destroyed, in-flight session dies. CC and Codex survive, because they run in alternate screen, but everything rendering inline in the primary buffer looks like an idle shell to the current Muxy check.

Turns out isTerminalIdle() in the view only looks at the alt screen. TerminalOfflinePolicy.isIdle already has a hasRunningProcess term, but nothing in prod ever passed it — only the tests did.

So this mostly just wires the view through the policy. A pane counts busy unless its foreground process is a known shell (sh, bash, zsh, fish, …), read via the existing ghostty_surface_foreground_pid + proc_name. Unknown or unreadable name counts as busy — better to keep a pane awake than kill a live session on a bad read.

First attempt was comparing the foreground pid to the session leader instead of matching names. Doesn't work: Muxy spawns panes through /usr/bin/login, which owns the session, so even an idle shell reads as busy and the feature silently never fires. Name matching against a closed shell set is the version that actually works.

Tested on a dev build with the threshold at 60s: idle shell tab still goes offline (~80s), tab running omp stays alive well past it. scripts/checks.sh green on the pinned tool versions.

## Summary by CodeRabbit

* **New Features**
  * Improved terminal activity detection using process details and command start/finish signals.
  * Enhanced offline and idle-state decisions for interactive shells, running commands, and alternate-screen sessions.
  * Improved remote terminal stream attachment and cleanup.
  * Improved shell input tracking and support for terminal multiplexers.

* **Bug Fixes**
  * Prevented stale terminal output sessions from affecting current command status.
  * Added safer handling when process information is unavailable.

* **Tests**
  * Added coverage for process inspection, shell activity tracking, and offline-state detection.

Disclosure: implemented with AI assistance (Claude Fable 5). PR text is mine.


dergachoff:
@saeedvaziry nice, tracking shell activity seems better than my name-matching, should catch more backgrounded jobs 👍

PR #1002, opened 2026-08-03
Fix agent tab startup commands

chamhaw:
## Summary
- Preserve pane startup commands across terminal tab snapshot and restore so command tabs run when materialized.
- Launch local agents through the user shell command name while keeping executable-path detection for availability.
- Preserve the existing POSIX-style startup wrapper for sh/bash/zsh-like shells and add fish-compatible local and remote startup wrappers.

## Shell compatibility
The existing POSIX-style startup script remains the default path for sh/bash/zsh-like shells. This PR adds fish-specific startup script generation because fish is not POSIX-compatible and does not reliably run the previous interactive `-i -c` startup path in the persistent session flow.

Remote terminal startup now enters a `/bin/sh` bootstrap before dispatching back to the remote `$SHELL`. This keeps the outer SSH command parseable when sshd invokes a fish login shell with `-c`, while still preserving the existing POSIX-style execution path for sh/bash/zsh-like shells.

This does not claim general support for every non-POSIX shell; it only covers the fish startup behavior validated here.

## Testing
- mise exec -- swift test --filter TerminalLaunchCommandTests
- mise exec -- swift test --filter SessionDaemonEndToEndTests
- mise exec -- scripts/checks.sh --fix

AI assistance: OpenAI GPT-5 Codex.

## Summary by CodeRabbit

* **New Features**
  * Terminal startup commands now preserve interactive mode, close-on-exit settings, and restoration preferences when tabs are restored.
  * Added Fish shell support for local and remote startup commands.
  * Agent launches now use available configured provider executables.

* **Bug Fixes**
  * Improved handling of unavailable agent executables and shell-specific startup behavior.

* **Tests**
  * Added coverage for startup restoration, one-time launches, and Fish/Zsh command execution.

### DIFF
--- a/Muxy/Models/Project/CommandTabRequest.swift
+++ b/Muxy/Models/Project/CommandTabRequest.swift
@@ -7,6 +7,7 @@ struct CommandTabRequest: Equatable {
     let command: String
     let closesOnCommandExit: Bool
     let directory: String?
+    let startupCommandRestoration: StartupCommandRestoration
 
     init(
         projectID: UUID,
@@ -14,7 +15,8 @@ struct CommandTabRequest: Equatable {
         name: String,
         command: String,
         closesOnCommandExit: Bool,
-        directory: String? = nil
+        directory: String? = nil,
+        startupCommandRestoration: StartupCommandRestoration = .restore
     ) {
         self.projectID = projectID
         self.areaID = areaID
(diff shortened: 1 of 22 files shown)


saeedvaziry:
image

This PR introduces bugs. please have a look

saeedvaziry:
image

the problem is not solved yet. also every time opening the plus icon's popover it takes time to load. I dont think this is efficient way of solving the problem honestly.

chamhaw:
Superseded by #1045, which replaces this conflicting branch with the focused shell-command launch fix and its follow-up review coverage.

PR #1029, opened 2026-08-06
Respect the Auto-copy setting in the selection-clipboard write path

### LINKED ISSUE #963: Terminal selection is copied even when Auto-copy is disabled
### Description

Even when `Auto-copy selected text` is disabled, selecting text in the terminal still automatically copies it to the clipboard. I expected selected text to be copied only when this option is enabled

Image

### Steps to Reproduce

1. Open Settings.
2. Go to Terminal.
3. Make sure `Auto-copy selected text` is disabled.
4. Select some text in the terminal.
5. Notice that the selected text is automatically copied to the clipboard even though the option is disabled.

### Muxy Version

Version 1.3.0-beta.873 (873)

### macOS Version

27.0 Beta (26A5388g)

YuriNachos:
When "Auto-copy terminal selection" is off, selecting text in a terminal still copied it behind the scenes. The mouse-up path already respected the preference and showed the "Copied" toast, but the separate path libghostty uses to write the selection clipboard ignored it entirely — so the setting wasn't actually the single source of truth it appeared to be.

This makes that ghostty-side write path honor the same setting: when a selection-clipboard write comes in and auto-copy is disabled (or simply unset, which reads as disabled), it is now a no-op. The standard clipboard is left alone, and the existing mouse-up behavior plus the toast are unchanged.

The decision is pulled out into a tiny pure helper (`shouldWriteSelectionClipboard`) so it can be covered directly by unit tests, alongside adapter-level tests that drive the real preference and pasteboard.

Closes #963.

Generated with the assistance of GLM-5.2.

## Summary by CodeRabbit

* **Bug Fixes**
  * Selection clipboard writes now respect the auto-copy setting.
  * Standard clipboard writes continue to work independently of the auto-copy preference.
  * Clipboard contents and non-text formats are preserved during selection clipboard operations.

* **Tests**
  * Added coverage for enabled, disabled, and unset auto-copy settings.
  * Added coverage confirming standard clipboard behavior and pasteboard cleanup.

### DIFF
--- a/Muxy/Services/Terminal/GhosttyRuntimeEventAdapter.swift
+++ b/Muxy/Services/Terminal/GhosttyRuntimeEventAdapter.swift
@@ -137,6 +137,10 @@ final class GhosttyRuntimeEventAdapter: GhosttyRuntimeEventHandling {
 
     func writeClipboard(location: ghostty_clipboard_e, content: UnsafePointer<ghostty_clipboard_content_s>?, len: UInt) {
         guard let content, len > 0 else { return }
+        if location == GHOSTTY_CLIPBOARD_SELECTION {
+            let autoCopyEnabled = UserDefaults.standard.bool(forKey: GeneralSettingsKeys.autoCopyTerminalSelection)
+            guard Self.shouldWriteSelectionClipboard(settingEnabled: autoCopyEnabled) else { return }
+        }
         let buffer = UnsafeBufferPointer(start: content, count: Int(len))
         for item in buffer {
             guard let dataPtr = item.data else { continue }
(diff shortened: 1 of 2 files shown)


saeedvaziry:
@YuriNachos would you check and address the review? also please merge main so it runs the actions again (actions dont work because of the Github incident yesterday)

YuriNachos:
Rebased onto `main` (branch was already current with it) and the workflow re-ran: the `Checks` job is green now (passing, ~4 min) and CodeRabbit's review found no actionable comments. GitHub action runs are healthy again after yesterday's incident. Ready whenever you are — let me know if you'd like any changes.

PR #1036, opened 2026-08-08
fix: remove orphaned worktree dir when its main repo is missing

### LINKED ISSUE #808: Cannot remove worktree when main repository has been deleted: "'.git' is not a .git file, error code 7"
### Description

I had a worktree checked out at ~/Code/gpmp/grupos-inteligentes/api-renovate-fix. The main repository it belonged to (~/Code/gpmp/grupos-inteligentes/api) was deleted outside of Muxy beforehand. When I tried to remove the worktree through Muxy, it failed with:

> fatal: validation failed, cannot remove working tree: '/Users/pedromaia/Code/gpmp/grupos-inteligentes/api-renovate-fix/.git' is not a .git file, error code 7

I expected Muxy to detect the orphaned worktree and offer to clean it up (e.g. remove the directory and prune the stale worktree reference), or at least show a clearer message explaining that the main repository is missing.

### Steps to Reproduce

1. Open Muxy
2. Have a worktree whose main repository has been deleted/moved outside Muxy
3. Try to remove the worktree from Muxy
4. Observe error code 7 and the worktree remains

### Muxy Version

1.0.0

### macOS Version

26.3.1

YuriNachos:
## Summary

Fixes #808. When a project's main repository directory is deleted outside Muxy, removing one of its worktrees from Muxy used to throw and leave the orphaned worktree directory on disk forever. `cleanupOnDisk` ran the git `removeWorktree` call (a throwing `try`) before deleting the worktree directory, so when the main repo was gone git failed and aborted before the directory was ever removed. Now the git call is skipped when the local main repo no longer exists, and the orphaned directory is removed directly.

## Changes

- `WorktreeStore.cleanupOnDisk(worktree:repoPath:...)`: guard the `GitWorktreeService.removeWorktree` call on `context.isRemote || FileManager.default.fileExists(atPath: repoPath)`, so a local repo that was deleted out-of-band skips git (nothing left to unregister) and falls through to `fileOps.removeItem`. Remote path and real git errors (repo present) are unchanged.
- Added `cleanupRemovesOrphanedWorktreeWhenRepoIsMissing` to `GitWorktreeServiceRemoveTests` (reuses the `TempGitRepo` fixture): commit → add worktree → delete the repo dir → `cleanupOnDisk` must not throw and the worktree directory must be gone.

## Testing

## Screenshots

No UI changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Worktrees are preserved when their main repository is unavailable, while tracked entries are removed cleanly.
  * Cleanup accurately reports whether files were removed and avoids deleting worktree files when the repository is missing.

* **Documentation**
  * Clarified that `dirRemoved: false` means worktree files remain in place.

* **Tests**
  * Added coverage for missing repositories, reused paths, dirty worktrees, and project-level cleanup.


YuriNachos:
Rebased onto `main` (`f18a9aa`) and resolved the conflict with #1039's rewritten `WorktreeStore.cleanupOnDisk`.

The #808 guard — `context.isRemote || FileManager.default.fileExists(atPath: repoPath)` — now wraps the `GitWorktreeService.shared.removeWorktree(...)` call itself, keeping #1039's `force` / `timeout` / `OperationDeadline` / `Bool?` machinery intact. When the main repo dir is gone (local), it skips the doomed git call and removes the orphaned worktree dir directly via `context.fileOps.removeItem`, so the post-call `directoryRemoved` check still resolves to `true`.

`scripts/checks.sh` is green, and the #808 regression test (`cleanupRemovesOrphanedWorktreeWhenRepoIsMissing`) passes against the new API.

YuriNachos:
I noticed #1039 (bounded worktree-removal timeouts) land in the same part of WorktreeStore, which reminded me to check in on this one — it's been open a few days without a human look yet. Does the direct-deletion fallback for orphaned worktree directories seem reasonable to you, or would you rather see the CodeRabbit variant where genuine removal errors propagate instead of being ignored? Whichever direction you prefer, I'm glad to make the change and re-run the suite.

saeedvaziry:
@YuriNachos TBH, deletions are always sensitive. One example, in the past we had PR merge deleting the existing worktree and its branch because they are merged but some users found it unwanted deletion. Although doesnt make sense to keep the orphan worktree dirs, but deleting them without user approval might not be a good idea

YuriNachos:
That's a fair concern, and the past incident you describe is exactly the failure mode this should avoid — thank you for the context.

I'd rather not defend silent deletion. Two ways I can reshape this, and I'm happy to do either:

1. **Detect + report only**: the fallback notices an orphaned worktree directory but never removes anything — it logs a warning (and, where it fits, surfaces the path) so the user can delete it deliberately.
2. **Opt-in cleanup**: keep the removal, but only behind an explicit user action/confirmation rather than an automatic fallback.

If neither fits the direction you want, I'm equally fine closing this and letting the detection discussion happen in an issue first. Which would you prefer?

PR #1037, opened 2026-08-09
fix: parse the minus key as a literal key in KeyCombo

YuriNachos:
**Suggested title:** `fix(keyboard): parse the minus key as a literal key in KeyCombo`

## Summary

`KeyCombo(parsing: "cmd+-")` returned `nil`, so the minus key could never be bound through the
shortcut parser. The parser split its token string on both `+` and `-`; since `-` is also a real
key (`kVK_ANSI_Minus` / keypad minus), a literal minus was eaten as a separator. This splits on
`+` only, so `cmd+-` now parses to `KeyCombo(key: "-", command: true)`.

## Changes

- `KeyCombo.init(parsing:)`: split tokens on `+` only instead of `+` and `-`.
- `KeyComboParsingTests`: add `parsesMinusKey` (covers `cmd+-`, `cmd+shift+-`, `ctrl+-`) and extend the `tokenString` round-trip list with `cmd+c` and `cmd+-`.

## Testing

- [ ] Tested manually on macOS

Added parsing tests fail at HEAD (RED: `cmd+-` → nil) and pass after the fix (GREEN). Full gate
(format → lint → build-tests → test) is green.

## Screenshots

No UI changes.

---

**Note for review (not part of the PR text):** the plus key (`cmd++`) still cannot round-trip —
`tokenString` serialises a cmd+plus combo as `"cmd++"`, and splitting that on `+` consumes both
plus signs. That is a separate, format-level problem (the `+` key collides with the `+` separator)
and is intentionally out of scope here; only the `-` separator was removed.

Code written with assistance from GLM-5.2.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shortcut parsing so the minus key is recognized correctly when used with modifier keys.
  * Shortcut descriptions containing `-` now remain intact during parsing.

* **Tests**
  * Added coverage for minus-key shortcuts and token round-tripping.

### DIFF
--- a/Muxy/Models/Keyboard/KeyCombo.swift
+++ b/Muxy/Models/Keyboard/KeyCombo.swift
@@ -162,7 +162,7 @@ struct KeyCombo: Codable, Equatable, Hashable {
     init?(parsing description: String) {
         let tokens = description
             .lowercased()
-            .split(whereSeparator: { $0 == "+" || $0 == "-" })
+            .split(whereSeparator: { $0 == "+" })
             .map { $0.trimmingCharacters(in: .whitespaces) }
             .filter { !$0.isEmpty }
         guard let rawKey = tokens.last else { return nil }
--- a/Tests/MuxyTests/Services/Extensions/ExtensionShortcutStoreTests.swift
+++ b/Tests/MuxyTests/Services/Extensions/ExtensionShortcutStoreTests.swift
@@ -256,6 +256,13 @@ struct KeyComboParsingTests {
         #expect(KeyCombo(parsing: "cmd+return") == KeyCombo(key: KeyCombo.returnKey, command: true))
     }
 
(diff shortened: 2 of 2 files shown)


saeedvaziry:
@YuriNachos have you tested this?

YuriNachos:
Yes — there is a unit test for it and CI is green on this branch.

Without the change, `cmd+-` split into an empty token list because `-` was treated as a separator, so parsing returned `nil`. With the change, the `-` survives as the literal key. I covered it directly and via round-trip:

- `parsesMinusKey`: `cmd+-`, `cmd+shift+-`, `ctrl+-` all parse to `key: "-"` with the expected modifiers.
- `tokenStringRoundTrips` now also includes `cmd+-`, so the parsed combo round-trips back through `tokenString`.

`swift test` passes locally and CI is green here.

PR #1039, opened 2026-08-09
Bound worktree removal timeouts

### LINKED ISSUE #1017: git.worktree.remove: unbounded on local workspaces, uncancellable, and serialized app-wide
`git.worktree.remove` has no timeout on local workspaces, cannot be cancelled, and serializes every other worktree read while it runs. On a JS monorepo checkout that combination is a multi-minute stall with no way for the caller to tell success from a hang.

### Measured

A worktree checkout of a Rails + JS monorepo here is **1.7 GB / 214,769 inodes** (node_modules, the package-manager store, test caches). Unlink throughput on an idle SSD is **~4,150 files/s** (20,000 files removed in 4.8 s), so one `worktree.remove` takes **~50 s** — and longer under load.

### What that runs into

1. **No timeout on the local path.** `GitProcessRunner.runGit` wraps only the SSH branch (`SSHCommandRunner.withTimeout`, 60 s); the local branch returns `try await runProcess(...)` unbounded. `WorktreeTeardownProcess` (the user `$SHELL` commands from `.muxy/worktree.json`) has none either. For comparison, `exec` and `http.fetch` both default to 30 s. A wedged local git call is a JS promise that never settles.

2. **No cancel.** `muxy.execAsync` (#865) is the only cancellable native path and it isn't available on webview pages, so an extension calling `git.worktree.remove` has no way to abandon it — only to stop waiting.

3. **Everything queues behind it.** `GitWorktreeService` is a single `actor` with a `static let shared` covering `listWorktrees`, `hasUncommittedChanges`, `addWorktree`, `removeWorktree` and `pruneWorktrees` app-wide, so a 50 s removal in one project stalls worktree reads in every other project, for every extension and for Muxy's own sidebar. `MuxyAPI.Git` is additionally `@MainActor` — #848 ("Move worktree teardown/deletion logic to backend to avoid UI freezing") was closed, but that isolation looks unchanged.

4. **The result says nothing.** `worktree.remove` returns `Result<Void, APIError>`. `WorktreeStore.cleanupOnDisk` runs `git worktree remove --force`, then `prune` (in `try?`, so failures are swallowed), then `fileOps.removeItem`. An interrupted unlink leaves a directory git has already forgotten and `git worktree prune` will never reclaim, and the caller cannot tell — related to #808. I have one such orphan on disk, down to a single empty subdirectory.

### Suggestions, smallest first

1. Accept a `timeoutMs` on the local path of `GitProcessRunner.runGit`, defaulting like `exec`, and apply it to `WorktreeTeardownProcess` too.
2. Settle the JS promise once `git worktree remove` returns — git has forgotten the worktree at that point, which is the user-visible outcome — and let `fileOps.removeItem` finish afterwards. Report in the result whether the directory survived.
3. Emit a `worktree.removed` event (e.g. `{ projectID, path, dirRemoved }`), so a caller doesn't have to poll for completion.
4. Give `GitWorktreeService` per-repo isolation (an actor keyed by repo path), or move the unlink off the actor, so one slow removal doesn't block reads elsewhere.

### Extension-side workaround, for context

In a sidebar extension I now give the removal a 300 s budget and, on timeout, poll `git.worktrees()` until the path is gone before deciding it succeeded. That works only because the shared actor makes the list read queue behind the removal — which is the bug being leaned on. A `[ -d ]` shell check catches the leftover directory, and removals are serialized extension-side so a queued one doesn't spend its budget waiting for the previous one.

pedrommone:
## Summary

- add bounded local Git and teardown subprocess execution with process-group cancellation
- apply one removal deadline across consent-gated path resolution, cleanup, and verification
- return structured `{ path, dirRemoved }` results and forward `timeoutMs` through every extension bridge
- preserve project isolation, dirty worktrees, remote path normalization, and uncertain residual paths
- document the extension API and add lifecycle, bridge, path, and cleanup regression coverage

## Testing

- `scripts/checks.sh --fix` (format, lint, build, 3,240 tests)
- targeted worktree/process regression suite (51 tests)
- `git diff --check`

Fixes #1017

AI assistance: OpenAI GPT-5.6 Sol

## Summary by CodeRabbit

* **New Features**
  * Added timeout controls for worktree removal, teardown, Git commands, and file checks.
  * Worktree removal now reports the resolved path and directory cleanup status.
  * Added support for remote worktree paths and improved path validation.
  * Processes now stop cleanly when canceled or timed out.

* **Bug Fixes**
  * Improved handling of incomplete cleanup, symbolic links, partial output, and dirty worktrees.
  * Git errors now provide clearer timeout and cancellation details.

* **Documentation**
  * Updated guidance for timeout limits, cleanup status, and removal behavior.


pedrommone:
Addressed the review in `251db31`:

- Anchored untracked path resolution and removal to `project.path`.
- Made Git output reads cancellable without closing handles concurrently with active readers.
- Moved `OutputReader` callbacks outside its lock while preserving callback order.
- Moved parent-directory cleanup off the main actor.
- Added real-repository timeout coverage, raised the teardown test timeout to 0.5 seconds, covered omitted bridge timeouts, and documented timeout errors.

Review items intentionally not changed:

- `SubprocessRunner` already has the requested timeout/cancellation completion path: timeout terminates the process group, deferred reaping completes, and `didTerminate()` forcibly finishes both readers when a termination reason exists. This does not depend on descendant pipe EOF.
- Reaping an `.exited` leader immediately would skip terminating surviving process-group descendants, which can keep output pipes open.
- Replacing the bounded local presence subprocess with `lstat` plus `AsyncTimeout.run` would weaken the hard timeout: the task group waits for a blocking syscall that ignores cooperative cancellation.
- The shared actor remains reentrant across awaited process operations, so it does not serialize unrelated repositories for the duration of a removal.
- Docstrings were not added because this repository explicitly prohibits code comments.

Validation: `scripts/checks.sh --fix` passes formatting, strict lint, build, and all 3,240 tests; the focused timeout/removal/bridge suite passes all 51 tests.

PR #1066, opened 2026-08-18
feat: option to create independent terminal tabs on split

dzulfikar08:
## Summary

Adds a **"Create independent terminal tabs"** setting under Terminal Settings. When enabled, Command-D / Command-Shift-D open the new terminal as its own top-level tab group — side-by-side in one screen, each pane with its own tab strip — instead of the current in-tab split that groups the new pane under the active tab. The default is unchanged.

## Why

The current split groups the new terminal under the active tab in the tab strip. Some users prefer the older behavior where each split pane behaves as an independent terminal tab. Making it a setting keeps the existing default for everyone while allowing the older workflow.

## Checks

- `swiftformat --lint` passes on all changed files
- `swiftlint lint --strict` passes on all changed files
- `swift build -c release` passes
- `swift test` — full suite passes

## AI contribution notice

As required by the contribution guidelines, this work was developed with the assistance of DeepSeek V4 flash. The author reviewed and verified the code.

### DIFF
--- a/Muxy/Models/Preferences/TerminalSplitPreferences.swift
+++ b/Muxy/Models/Preferences/TerminalSplitPreferences.swift
@@ -1,0 +1,15 @@
+import Foundation
+
+enum TerminalSplitPreferences {
+    static let independentTabsKey = "muxy.terminalSplit.independentTabs"
+    static let defaultIndependentTabs = false
+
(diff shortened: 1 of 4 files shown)


saeedvaziry:
@dzulfikar08 if the PR ready to be reviewed please mark it so

PR #1082, opened 2026-08-23
Fix mobile scrollback replay on pane takeover

wjchoi87:
## Summary

Follow-up to the mobile scrollback feature
  (2bd…/23279750 "Buffer terminal raw output for mobile scrollback"). Pane
  takeover replayed the raw history buffer as-is, which broke in three ways:

  - The ring trim discarded bytes at arbitrary offsets, so the replay head could
    start mid-escape-sequence / mid-UTF-8 with no terminal state established;
    long buffers rendered as literal text or garbage in the mobile emulator.
  - Alternate-screen frames (vim/less/opencode redraws) were stored whole,
    bloating the replay and polluting the main-screen scrollback.
  - The client re-entering a pane replayed the full history on top of the
    existing terminal without resetting it, accumulating duplicates.

  This PR sanitizes the replay server-side and documents the contract; the
  mobile client resets its terminal before consuming the replay (separate
  muxy-mobile change).

## Changes

Server (this repo)

  - Add `TerminalStreamSequence` (MuxyShared): shared escape/OSC/CSI and
    alternate-screen sequence scanner; `SessionReplayBuffer` now delegates to
    it with no behavior change (desktop reattach unaffected).
  - Add `MobileScrollbackBuffer` (ring buffer):
    - keeps alternate-screen enter/leave markers only, drops TUI frames, so
      history holds shell output, not full-screen app redraws
    - aligns the replay head to a line boundary after trimming and drops
      incomplete trailing escape/UTF-8 sequences
    - prepends a state-reset prefix (`ESC[0m ESC(B ESC[2J ESC[H`) so clients
      replay from a known state
    - appends in place (no full-buffer copy per write) and trims/shrinks when
      the configured cap changes
  - Switch `RemoteTerminalStreamer` to the new buffer.
  - Update `docs/remote-server` (events, methods): takeover now replays the
    sanitized history before the snapshot; clients must reset (screen +
    scrollback) before consuming the replay.

  Mobile (muxy-mobile, separate PR)

  - `terminalHtml.ts`: takeover now always runs `term.reset()` + `term.clear()`
    before writing the replay, removing duplicate history across re-entries.

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added terminal scrollback retention for remote panes with configurable capacity and trimming.
  * Rejoining or taking over a pane now replays retained terminal history before the current snapshot and live output.
  * Replay data is sanitized for safe UTF-8 and escape-sequence boundaries, excluding alternate-screen content.
  * Invalid scrollback capacity settings are rejected without overwriting existing values; null settings reset to the default.

* **Documentation**
  * Updated remote server documentation to describe takeover replay behavior and terminal-clearing requirements.


pedrommone:
it happens a lot!

PR #1092, opened 2026-08-26
feat: add Global Workspace

vipinchan:
## Summary

Adds an optional Global Workspace: a dedicated, globally accessible full Muxy workspace that can be shown or hidden independently from the main app window.

Unlike Quick Terminal, which provides a lightweight single-terminal surface, Global Workspace hosts the full `MainWindow` experience with independent workspace state, including projects, tabs, panes, sidebar, and related workspace UI, together with overlay-fullscreen presentation.

Global Workspace is configured from **Settings → Global Workspace**.

### Behavior

- Show or hide Global Workspace from any app using:
  - Double Command
  - Double Control
  - Double Option
  - a custom global key combination
- Configurable modifier double-tap interval
- Optional toggle-to-hide behavior
- Restores focus to the previously active application when hidden
- Keeps Global Workspace UI and workspace state independent from the main Muxy window
- Supports normal-window and overlay-fullscreen presentation
- Preserves overlay-fullscreen state across hide/show
- Can be presented across macOS Spaces and over fullscreen apps

### Quick Terminal vs. Global Workspace

Quick Terminal remains the lightweight, single-terminal workflow for fast commands.

Global Workspace provides a separate full-workspace workflow for users who want projects, tabs, panes, sidebar, and the rest of the Muxy workspace available globally without changing the main-window workspace state.

The two features share shortcut infrastructure where appropriate while retaining separate controllers, state, and lifecycles.

### Shortcut integration

This implementation is rebuilt on Muxy’s current shortcut architecture rather than carrying forward the original standalone hotkey implementation:

- reusable double-modifier backend for modifier-only gestures
- existing Carbon hotkey backend for custom global key combinations
- separate Quick Terminal and Global Workspace shortcut services
- existing Quick Terminal behavior remains unchanged

### Implementation

- dedicated `GlobalWorkspaceController` and workspace panel
- dedicated `AppState` and workspace persistence for Global Workspace
- shared application stores and services rather than duplicating application source-of-truth state
- reuse of `MainWindow` for the full workspace UI
- top-level **Settings → Global Workspace** configuration
- `settings.json` validation and persistence support
- overlay-fullscreen presentation that remains on the current Space
- overlay-fullscreen state preserved across hide/show

## Testing

Additional coverage and validation includes:

- unit coverage for Global Workspace preferences and shortcut behavior
- double-modifier backend tests
- Settings catalog and settings persistence coverage
- `settings.json` validation and persistence coverage
- global show/hide from another application
- focus restoration after hiding
- Double Command, Double Control, and Double Option triggers
- custom global shortcuts
- independent Global Workspace and main-window workspace state
- normal-window presentation
- overlay-fullscreen presentation
- overlay-fullscreen state preserved across hide/show
- full workspace interaction including projects, tabs, panes, sidebar, and terminal
- GitHub Actions full check suite, including formatting, strict linting, test builds, and the complete test suite

## Screenshots

The recording shows Global Workspace being opened from another application, hidden with focus restored to the previous application, switched to overlay fullscreen, and shown again with the fullscreen state preserved.

muxy

## Summary by CodeRabbit

* **New Features**
  * Added Global Workspace, a floating workspace available across apps and spaces.
  * Added configurable double-tap modifier and custom shortcut triggers.
  * Added controls for timing, visibility, full-screen behavior, and input-monitoring access.
  * Added dedicated settings with persisted configuration and shortcut conflict validation.

* **Bug Fixes**
  * Improved window identity, focus restoration, layout persistence, and full-screen notification handling.

* **Documentation**
  * Added setup and usage guidance for Global Workspace, shortcuts, permissions, and visibility controls.

### DIFF
--- a/Muxy/Models/Layout/AppLayoutStore.swift
+++ b/Muxy/Models/Layout/AppLayoutStore.swift
@@ -6,11 +6,13 @@ final class AppLayoutStore {
     static let shared = AppLayoutStore()
 
     private let defaults: UserDefaults
+    private let storageKey: String
     private(set) var layout: AppLayout
 
-    init(defaults: UserDefaults = .standard) {
+    init(defaults: UserDefaults = .standard, storageKey: String = AppLayout.storageKey) {
         self.defaults = defaults
-        let stored = defaults.string(forKey: AppLayout.storageKey)
+        self.storageKey = storageKey
+        let stored = defaults.string(forKey: storageKey)
         layout = stored.flatMap(AppLayout.init(rawValue:)) ?? AppLayout.defaultValue
     }
 
@@ -19,7 +21,7 @@ final class AppLayoutStore {
     func set(_ layout: AppLayout) {
         guard self.layout != layout else { return }
         self.layout = layout
-        defaults.set(layout.rawValue, forKey: AppLayout.storageKey)
+        defaults.set(layout.rawValue, forKey: storageKey)
     }
 
     func toggle() {
--- a/Muxy/MuxyApp.swift
+++ b/Muxy/MuxyApp.swift
@@ -90,6 +90,13 @@ struct MuxyApp: App {
                     .environment(ExtensionSettingsStore.shared)
                     .preferredColorScheme(MuxyTheme.colorScheme)
                     .onAppear {
+                        GlobalWorkspaceController.shared.configure(
+                            projectStore: projectStore,
+                            worktreeStore: worktreeStore,
+                            projectGroupStore: projectGroupStore,
+                            remoteDeviceStore: remoteDeviceStore,
+                            browserStores: (profiles: browserProfileStore, history: browserHistoryStore)
+                        )
                         startDeferredServicesIfNeeded()
                         startWorktreeAutoRefreshIfNeeded()
                         NotificationStore.shared.appState = appState
@@ -844,9 +851,12 @@ struct WindowConfigurator: NSViewRepresentable {
         let v = NSView()
         DispatchQueue.main.async {
             guard let w = v.window else { return }
-            w.identifier = ShortcutContext.mainWindowIdentifier
-            if Self.closeDuplicateMainWindow(w) {
-                return
+            let isGlobalWorkspace = ShortcutContext.isGlobalWorkspaceWindow(w)
+            if !isGlobalWorkspace {
+                w.identifier = ShortcutContext.mainWindowIdentifier
+                if Self.closeDuplicateMainWindow(w) {
+                    return
+                }
             }
             w.titlebarAppearsTransparent = true
             w.titleVisibility = .hidden
@@ -858,7 +868,9 @@ struct WindowConfigurator: NSViewRepresentable {
             Self.repositionTrafficLights(in: w)
             Self.hideTitlebarDecorationView(in: w)
             Self.neutralizeSafeAreaInsets(in: w)
-            Self.interceptCloseButton(in: w, coordinator: context.coordinator)
+            if !isGlobalWorkspace {
+                Self.interceptCloseButton(in: w, coordinator: context.coordinator)
+            }
             context.coordinator.observe(window: w)
         }
         return v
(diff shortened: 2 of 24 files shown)


vipinchan:
This replaces #893. I rebuilt the feature on top of the current `main` branch and Muxy’s current Quick Terminal / shortcut architecture, while keeping Global Workspace as a separate full-workspace feature. The branch has also been rebased to the current upstream base and the full Checks workflow passes.

vipinchan:
@saeedvaziry this is ready for review. This is the replacement for #893, rebuilt on the current shortcut architecture and current `main`.

saeedvaziry:
@vipinchan would you please address the reviews?

vipinchan:
Hi @saeedvaziry, all review feedback has been addressed in commits 9166215 and c72570a. All threads are resolved, CodeRabbit is passing, and CI is green on my fork (upstream is awaiting approval).
As a longtime iTerm2 power user, I rely on this feature heavily in my daily workflow, so I'm excited to see it get added. The PR is ready for re-review. Thank you!

saeedvaziry:
@vipinchan can you please what is the purpose of the global workspace?

Muxy is not efficient for multi-window. if you try this PR's feature yourself for a while you will see how laggy and inefficient the app will become, other issues like quitting will quit all or changing sidebar will toggle the other the other way around and many many such issues.

Muxy currently is not ready for any multi-window feature and I wouldn't add such features to the app.

I can recommend coming back to this as a discussion once I am done with re-writing the full app in Rust.

We need to first enable Muxy from its core to support multiple windows.

vipinchan:
Thanks for the clarification, @saeedvaziry . Just to explain the actual use case a bit better:

I’m not really looking for general multi-window support. What I want is simply to be able to summon Muxy from anywhere with a global shortcut and continue working immediately. I currently use Double Command for this, and hiding it returns focus to the previous app. The recording in the PR shows the workflow. This has been useful enough that I moved from iTerm2 to Muxy and now use it this way every day.

I understand your concern about the current implementation. Technically it does create a second full workspace/window, but that wasn’t the goal itself. I went with this approach mainly because reusing the existing `MainWindow` directly turned out to be much more invasive around window lifecycle, Spaces, fullscreen, and show/hide behavior.

I haven’t personally noticed any obvious performance issues so far, but I understand that my own usage doesn’t replace a broader architectural or long-term stability assessment. 

This feature is off by default and only meant for this global summon workflow.

saeedvaziry:
Ah now I see your idea.

I have that need as well but maybe I give you a workaround.

I use Raycast in my mac and in Raycast you can set a shortcut for each application. so in my system I hit cmd+shift+space and Muxy being focused immediately and doesnt matter where I am.

vipinchan:
@saeedvaziry thanks for the reply. I understand the Raycast approach — it solves the “quickly focus Muxy” part. What I’m trying to achieve is a bit more than that: a really seamless way to move between Muxy and whatever I’m currently working on.

I’d actually suggest trying the workflow in the PR. Instead of switching to the normal Muxy window, it temporarily brings up a floating Muxy workspace on top of the current app. When I’m done, I trigger it again and it disappears, taking me straight back to the app I was using before.

For me, that’s the main difference: Raycast is “quickly switch to Muxy,” while this is more like “seamlessly jump into Muxy and back again” without breaking the flow.

saeedvaziry:
@vipinchan I opened the app from the PR. set double-command to trigger. when I hit double command, it opens a new Muxy window and overal I have 2 Muxy windows which causing both to be laggy and not functioning

vipinchan:
Thanks for trying it, @saeedvaziry. That makes sense.

It works well for my setup, but if the second workspace makes both windows laggy, I agree it’s probably not a good fit for upstream right now.

I’ll keep it in my fork for personal use, and I hope the Rust version can eventually support this kind of interaction.

Thanks again for testing it.

saeedvaziry:
@vipinchan thatnks for contributing anyway.

I have in my roadmap to support multi windows in Rust version

PR #960, opened 2026-07-27
fix: prevent SF Symbol 0×0 rasterization crash on macOS 27

Bono2007:
Sizes launch-path SF Symbols with explicit frames instead of font metrics, which collapse to 0×0 during the pre-visible layout pass and trigger the CoreUI recursive rendition-lock crash on macOS 27.0.

Fixes #941. Same root cause/fix as manaflow-ai/cmux#5670

## Summary by CodeRabbit

* **Style**
  * Improved consistency and sizing of system icons across browser, terminal, project, extension, workspace, settings, and notification interfaces.
  * Updated icons to scale and fit more reliably within their designated UI areas.
  * Preserved existing interactions, labels, colors, and accessibility behavior.

Issue #861, opened 2026-07-04
Regression: infinite LazyVStack layout cycle hangs app on 1.3.0-beta.806 (same signature as #775)


deniskern:
## Description

The app hangs again with an infinite SwiftUI layout cycle in `LazySubviewPlacements` — same signature as #775, which was fixed in #806. It reoccurs on **1.3.0-beta.806**, which already contains that fix, so this looks like a regression — possibly reintroduced with the new sidebar layout (#828, merged 2026-06-27; my hangs started 2026-07-02).

Main thread is unresponsive for 130+ seconds at 100% CPU until macOS marks the app as Not Responding and I have to force quit. Happening daily.

## Environment

- Muxy 1.3.0-beta.806 (806)
- macOS 26.5 (25F71), Apple Silicon (Mac17,9, 48 GB)

## Evidence

- 4 `.hang` reports (3× 2026-07-02, 1× 2026-07-03) in `/Library/Logs/DiagnosticReports/`
- 2 `cpu_resource.diag` reports: "90 seconds cpu time over 90 seconds (100% cpu average), exceeding limit of 50% cpu"
- Latest hang: `Duration: 133.12s`, all 11 samples in the same stack
- Every session for weeks ends with `PREVIOUS SESSION ENDED UNCLEANLY` in diagnostics.log

## Heaviest main-thread stack (trimmed to the distinctive frames)

```
GraphHost.flushTransactions()
specialized GraphHost.runTransaction(_:do:id:)
AG::Subgraph::update(unsigned int)
AG::Graph::UpdateStack::update()
specialized LazySubviewPlacements.updateValue()
LazyLayoutViewCache.commitPlacedSubviews(from:to:wasCancelled:context:containingSize:)
_LazyLayoutViewCache.initialPlacement(newIndex:newPlacedSubviews:oldPlacedSubviews:wasInsertedToSubviews:context:)
_LazyLayoutViewCache.withPlacementData<A>(_:)
closure #1 in _LazyLayoutViewCache.finalPlacement(...)
protocol witness for LazyLayout.finalPlacement(...) in conformance LazyVStackLayout
protocol witness for LazyLayout.initialPlacement(...) in conformance LazyVStackLayout
specialized LazyStack<>.initialPlacement(...)
Array<A>.motionVectors(closestTo:in:avoiding:distance:)
AGGraphGetValue
AG::Graph::index_of_input_slow(AG::Node&, AG::InputEdge::Comparator)
```

[...]

PR #1119, opened 2026-09-17
fix: stop the macOS 27 launch crash from 0×0 SF Symbol raster targets

### LINKED ISSUE #941: Crash on launch on macOS 27.0 beta — SF Symbol rasterized at 0×0 during initial window layout (CUINamedVectorGlyph assertion)
Summary

Muxy is killed on launch on macOS 27.0 beta (build 26A5388g), before any window appears. Reproduced on 1.2.1 stable and 1.3.0-beta.860, with a completely fresh install and empty state. The same builds run fine on macOS 26.6.

The crash is not a recursive-lock abort (an earlier guess). The real cause is a CoreUI assertion that macOS 27 added: a SwiftUI SF Symbol is rasterized at a target size of 0×0 during the window's initial (interpolated) layout, and CoreUI now aborts instead of tolerating it.

Environment

- macOS 27.0 beta, build 26A5388g (Apple Silicon / arm64)
- Muxy 1.2.1 and 1.3.0-beta.860 — both crash identically
- Works on the same machine under macOS 26.6

Exact assertion

*** Assertion failure in -[CUINamedVectorGlyph _rasterizeImageUsingScaleFactor:forTargetSize:renderingMode:colorResolver:], CUINamedVectorGlyph.m:2599
Invalid parameter not satisfying: targetSizeInPoints.width>0 && targetSizeInPoints.height>0

The process embeds a crash handler, so it exits with status 6 and no system .ips report is generated — hard to see without a debugger. Captured with a local debug build under lldb.

Backtrace (key frames)

CoreUI      -[CUINamedVectorGlyph _rasterizeImageUsingScaleFactor:forTargetSize:renderingMode:colorResolver:]   ← asserts, targetSize = (0, 0)
CoreUI      -[CUINamedVectorGlyph image]
RenderBox   RB::Symbol::Presentation::template_image()
SwiftUICore _ShapeStyle_RenderedShape.renderVectorGlyph(...)
SwiftUICore InterpolatedDisplayList.updateValue()          ← interpolating from a zero-size prior state
SwiftUI     NSHostingView.layout()
AppKit      -[NSView _layoutSubtreeWithOldSize:]           ← oldSize = 0

Glyph at crash: valid _pointSize = 11, but forTargetSize: = (0.0, 0.0).

Root cause

During the window's first layout, NSHostingView.layout() produces an InterpolatedDisplayList interpolating from a previous size of 0. At the start of that interpolation every SF Symbol is resolved at size 0.
macOS 26.6 no-ops a 0-size glyph rasterization; macOS 27.0 added a targetSize > 0 assertiomic (the crash hits the first symbol in the initial frame — removing it just moves the crash to the next) and timing-sensitive (a debugger slows launch enough to miss the zero-size window).

Steps to reproduce

1. macOS 27.0 beta (26A5388g), Apple Silicon.
2. Install any current Muxy build and launch it.
3. Process aborts within ~1 s; no window.

Workarounds already ruled out

None prevented the crash (tested at full speed): .fixedSize(), .drawingGroup(), .transaction { $0.animation = nil } (image / status bar / MainWindow root); deferring applicationDidFinishLaunching; moving
UserDefaults.register(defaults:) earlier; neutralizing the .animation(.easeInOut, value:) Window behind a non-zero GeometryReader; NSAutomaticWindowAnimationsEnabled = false. Thezero-size render comes from SwiftUI's own NSHostingView interpolated first layout, below app view code.

Also worth filing with Apple

Fundamentally a CoreUI regression in the macOS 27.0 beta — identical code works on 26.6. Worth a Feedback Assistant report in parallel; may be fixed in a later seed with no app change.

guppylab:
Muxy 1.5.0 (945) dies about a second after launch on macOS 27.0 stable (26A428, Apple Silicon). No window ever shows up. This is the fix from #960, which got closed without merging and doesn't apply cleanly to main anymore, so I redid it.

The short version: SF Symbols that get their size from `.font(.system(size:))` don't have resolved font metrics yet during the window's pre-visible layout pass, so the raster target ends up 0×0. macOS 26.6 shrugged this off. macOS 27 asserts in `-[CUINamedVectorGlyph _rasterizeImageUsingScaleFactor:forTargetSize:renderingMode:colorResolver:]`.

What turns an assertion into a crash is the lock. The assertion fires while CoreUI is holding an `os_unfair_lock` inside `_performWithLockedRenditions:`. AppKit swallows the exception and never releases it, and then the next glyph render tries to take the same lock again:

```
_os_unfair_lock_recursive_abort
_os_unfair_lock_lock_slow
-[CUINamedVectorGlyph _performWithLockedRenditions:]
-[CUINamedVectorGlyph image]
RB::Symbol::Presentation::template_image() const
_ShapeStyle_RenderedShape.renderVectorGlyph(_:size:orientation:name:styles:layers:)
ShapeStyledDisplayList.updateValue()
NSHostingView.layout()
```

Same root cause as #941, except now it's a stable release instead of a beta. Several people hit it in #1114. The `muxy.quickTerminal.enabled: false` workaround suggested there didn't do anything for me, and neither did nuking `~/Library/Preferences/com.muxy.app.plist`.

## What I changed

Every SF Symbol on the launch path now gets an explicit frame instead of relying on font metrics:

```swift
// before
Image(systemName: "…").font(.system(size: S, weight: W))

// after
Image(systemName: "…").resizable().scaledToFit().frame(width: S, height: S)
```

183 occurrences, 73 files. It's the same transformation as #960 — that diff has rotted (files diverged, and `Muxy/Views/Terminal/RichInputSidePanel.swift` is gone entirely), so I reapplied it mechanically rather than trying to rebase it.

One thing worth a careful look during review: icons are now sized by frame rather than by font metric, so optical size can drift a little wherever a glyph's bounding box doesn't match its cap height.

## Testing

- `swiftformat --lint .` — clean, 0/636 files need formatting
- `swiftlint lint --strict` — 6 violations, but `origin/main` without this patch reports the exact same 6. My swiftlint is 0.65.1 and `.tool-versions` pins 0.57.1, which explains the noise. This branch adds nothing new.
- `scripts/checks.sh` — didn't get through it, it bails on the pinned-tool-version check
- `swift build` on main with the patch applied
- Manual, on macOS 27.0 (26A428), MacBookPro18,3: signed release build of v1.5.0 + this transformation launches and behaves normally. Same tree unpatched aborts with exit code 6 in roughly a second.

## Unrelated, but while I'm here

`scripts/build-release.sh` is also broken under Xcode 27. It looks for `Muxy_Muxy.bundle/ghostty-overrides/transparent-surface.conf`, but newer SwiftPM emits `Muxy_Muxy.bundle/Contents/Resources/ghostty-overrides/transparent-surface.conf`. Happy to open that as a separate PR if it's useful.

Fixes #941
Refs #1114, #960

## Summary by CodeRabbit

* **Style**
  * Updated icons throughout the app to use consistent, frame-based sizing and aspect-fit rendering.
  * Preserved existing icon choices, colors, layouts, accessibility labels, and interactions.
  * Improved visual consistency across browser, terminal, workspace, extension, project, composer, and settings interfaces.


guppylab:
> Drafted by Claude Opus 5 at my direction, as the contributing guide asks contributors to name the LLM. The measurements below were taken on my machine.

@coderabbitai is right about the weights, and the fix it suggested does not work either — I checked both.

I rendered a gear symbol at four weights three ways and compared the output:

- `.font(.system(size: 22, weight: w))` — weights render correctly, thin through black
- `.resizable().scaledToFit().fontWeight(w)` — all four render identically at the default weight, so `fontWeight` after `resizable()` is a no-op, exactly as the review said
- `NSImage(systemSymbolName:)` + `NSImage.SymbolConfiguration(pointSize:weight:)` — weights render correctly

So I replaced the approach instead of patching it. This branch now adds a small `SymbolIcon` view that builds the image through an explicit `NSImage.SymbolConfiguration`, carrying both the original point size and the original weight from each call site, with a cache keyed on name/size/weight. Every one of the 183 sites keeps the weight it had before: `arrow.down.circle.fill` is still `.bold` in `UpdateBadge`, the `ladybug.fill` pair in `DebugButton` is still `.semibold`, `globe` in `BrowserStartPage` is still `.light`, and so on.

It also avoids the crash for a more direct reason than the frame version did: `Image(nsImage:)` never enters the RenderBox symbol path, so `RB::Symbol::Presentation::template_image()` is never reached and CoreUI is never asked to rasterize a glyph whose metrics are still unresolved.

Checks on this revision:

- `swiftformat --lint .` — clean, 0/638 files need formatting
- `swiftlint lint --strict` — 6 violations, and `origin/main` without this branch reports the same 6, so nothing new
- `swift build` — clean
- Manual, macOS 27.0 (26A428), MacBookPro18,3: a signed release build of v1.5.0 carrying this change launches and runs; the same tree unpatched aborts with exit code 6 in about a second

One thing I found while testing that is worth knowing for #941, and that I will write up there in more detail: the 0×0 raster target is downstream of something else. In the affected process `NSImage(systemSymbolName:)` itself intermittently returns a degenerate 3.0×2.0pt image with no pixel data, and the first resolution of a given symbol name decides that name's fate for the life of the process — retrying never recovers it. A clean process on the same machine at the same moment resolves the same names correctly.

guppylab:
> Drafted by Claude Opus 5 at my direction, per the contributing guide's request to name the LLM.

Reverting the branch to the frame-sizing revision. The `NSImage.SymbolConfiguration` version I pushed an hour ago preserves the weights correctly, but it does not survive contact with the real app — I had only compiled it, not run it, which was my mistake.

Built on top of v1.5.0 and launched, it dies before any window appears:

```
*** Terminating app due to uncaught exception 'NSGenericException', reason:
'The window has been marked as needing another Layout Window pass, but it has already had
more Layout Window passes than there are views in the window.
<SwiftUI.AppKitWindow: 0x104c27760> 0x7af (1967) {{0, 64}, {1512, 885}} en'
...
-[NSWindow(NSDisplayCycle) _postWindowNeedsLayout]
-[NSWindow(NSConstraintBasedLayoutInternal) _layoutViewTree]
NSApplicationMain
```

I tried it twice: once with the image left at its natural size, once pinned with `.resizable().scaledToFit().frame(width:height:)`. Both loop. So swapping 183 SwiftUI symbol views for AppKit-image views perturbs the layout cycle somewhere, and the fix for the weight regression is not this straightforward. It may be related to #861, which reports an infinite layout cycle from a different trigger.

So the branch is back to the frame version, which I have running on macOS 27.0 (26A428) without the crash. The weight regression `@coderabbitai` identified is real and still unaddressed in this revision — worth knowing before merging, and I don't have a verified fix for it yet. Whoever picks this up: the weight cannot be restored with `.fontWeight()` after `.resizable()` (I checked, it is a no-op), so it needs either a different rendering path that doesn't disturb layout, or per-site frames tuned to match the old optical sizes.

guppylab:
> Drafted by Claude Opus 5 at my direction, per the contributing guide's request to name the LLM.

Pushed a fix for the weight regression. It is partial by design, and here is why.

Converting all 183 sites to the `NSImage.SymbolConfiguration` path is what broke the app earlier — the window enters an endless layout cycle and AppKit throws `NSGenericException` before anything appears. I confirmed the boundary by building four variants on top of v1.5.0 and launching each five times:

| variant | launches survived |
|---|---|
| all 183 sites via `.font(.system(size:weight:))` + explicit `.frame` | 0/5, layout cycle |
| all 183 sites via `SymbolIcon` (NSImage) | 0/5, layout cycle |
| all 183 via `.resizable().scaledToFit().frame` (this PR's base) | 5/5 |
| base + 10 sites via `SymbolIcon` | 5/5 |

So the layout cycle comes from converting the whole tree, not from either technique on its own, and a small number of conversions is safe.

This revision therefore keeps `.resizable().scaledToFit().frame` everywhere and routes only the ten sites `@coderabbitai` flagged through `SymbolIcon`, each with the weight it had on main:

- `BrowserStartPage`: `globe` `.light`, `magnifyingglass` `.medium`
- `UpdateBadge`: `arrow.down.circle.fill` `.bold`
- `DebugButton`: both `ladybug.fill` `.semibold`, metric icons `.medium`
- `IconButton`: `.semibold`
- `OpenProjectControl`: both icons `.semibold`
- `ExtensionWebViewPane`: `puzzlepiece.extension` `.light`

What this does not fix: the remaining 173 sites still render at the default weight. Most were `.medium` or `.regular`, where the difference is hard to see, but it is a real change and I am flagging it rather than claiming the regression is closed. Restoring those needs either a rendering path that preserves weight without disturbing layout, or per-site sizing tuned by eye — neither of which I can verify at that scale.

Verification on this revision: `swiftformat --lint` clean, `swiftlint --strict` at the same 6 pre-existing violations as main, `swift build` clean. The five-launch runs above were done on a v1.5.0 base, since running a build of current main against my own app data risks a one-way settings migration; the change applied to those files is identical.

saeedvaziry:
@guppylab can you please narrow the PR down to only the crash fix? I see unrelated changes which led to 74 file changes

guppylab:
> Drafted by Claude Opus 5 at my direction, per the contributing guide's request to name the LLM. The measurements below were taken on my machine, macOS 27.0 (26A428), MacBookPro18,3.

@saeedvaziry Narrowed it: the weight-regression work is out of this PR and now lives on `fix/macos27-sfsymbol-weight-regression`, to land separately on top of this one. #1119 is a single commit and 73 files.

The 73 files are all one mechanical substitution — no refactor, no unrelated changes:

```diff
- Image(systemName: "x").font(.system(size: S, weight: W))
+ Image(systemName: "x").resizable().scaledToFit().frame(width: S, height: S)
```

I also went looking for whether a smaller subset of call sites would do, and it does not. Two things I found while measuring that I think change the picture:

**1. The crash reproduces on a clean profile.** With an empty Application Support directory — a fresh install — the app still dies before any window appears, 3/3. It is not tied to my 20 projects / 21 tabs. If you cannot reproduce, I suspect it is the OS version rather than state.

**2. Subsetting the call sites does not converge.** The first symbol the renderer rasterizes is the one that dies, and fixing it just promotes the next one:

| build | launches survived |
|---|---|
| `main`, unpatched | 0/3 |
| only the two `cpu` sites (`ResourceUsageButton`) converted | 0/5 — next failure is `sidebar.left` |
| this PR, single commit | 5/5 on a clean profile, 3/3 on my real one |

Method: debug build, launched into an isolated `HOME`, 12s window per launch, classified by the exception signature in stderr rather than exit code.

**The failure is also not what the PR title says.** The real exception is:

```
NSInternalInconsistencyException: Invalid parameter not satisfying:
targetSizeInPoints.width>0 && targetSizeInPoints.height>0
Assertion failure in -[CUINamedVectorGlyph _rasterizeImageUsingScaleFactor:forTargetSize:renderingMode:colorResolver:],
CUINamedVectorGlyph.m:2599
```

Under `lldb` the glyph arrives with `alignment rect = {0, 0}` — e.g. `<CUINamedVectorGlyph 'cpu' @2x, 11-points, Semibold weight, Medium size, alignment rect = {Width = 0, Height = 0}>`. Every glyph in the process comes back that way, which is why no subset of sites helps. The exception is thrown from inside `-[CUINamedVectorGlyph _performWithLockedRenditions:]`, so it unwinds while holding an `os_unfair_lock`; the *next* symbol render on that thread then aborts with `BUG IN CLIENT OF LIBPLATFORM: Trying to recursively lock an os_unfair_lock`. That second signature is what the crash reports in `~/Library/Logs/DiagnosticReports` show, and it is a secondary symptom — the 0×0 raster target is the actual fault.

**One open lead, and a question for you.** A stock SwiftUI app on the same machine, same symbol, same point size and weight, resolves it correctly (`alignment rect = {12, 12}`), so this is something about our process state rather than a plain SF Symbols regression. I have one confirmed trigger: `MuxyTheme.colorScheme` initializes `GhosttyService.shared` lazily *during* a SwiftUI body evaluation. Forcing that init before the first render makes a probe view survive 3/3 where it otherwise dies 0/3. That alone is not sufficient with the real `MainWindow` — warming the singletons in `applicationWillFinishLaunching` did not fix it — so there is at least one more trigger I have not isolated.

These did **not** fix it, in case anyone else looks: warming all the app singletons early; the `.opacity(0)` on inactive `TerminalArea`s; removing `TerminalArea` entirely; `Color(nsColor:.withAlphaComponent)` reproduced standalone; the `ResourceUsageButton` view structure reproduced standalone.

So: do you want this workaround merged now and the root cause chased separately, or would you rather hold #1119 while I keep digging? If the real trigger turns out to be that lazy initialization, the fix is likely a few lines and this PR becomes unnecessary — but a clean install on macOS 27 currently cannot launch at all, which is why I would lean toward landing the workaround first.

